PR #378, opened 2015-08-09
Add systemd configuration

goetzk:
Because systemd is the future of init it seems sensible to have documentation
on how to use it. I've tested and deployed this configuration so will vouch for
its utility.

Its based on the file available at the following link from Matt-Deacalion.
https://github.com/Matt-Deacalion/systemd-django/blob/master/services/django-website.service
The copyright file in the repository says it is MIT licensed - I'm not convinced
that it is copyrightable but even if it is MIT is compatible with the licence of Tendenci.

### DIFF
--- a/docs/source/installation/remote.txt
+++ b/docs/source/installation/remote.txt
@@ -144,6 +144,55 @@ If any changes in the future are made to the file ``conf/local_settings.py``, th
 
     sudo service tendenci restart
 
+Setting up SystemD
+------------------
+
+On most distributions (Including recent Ubuntu systems) Systemd is the default
+init system.
+
+Systemd Unit files should be created in /etc/systemd/system/ with a name of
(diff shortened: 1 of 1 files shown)


jennyq:
Great addition for adding systemd setup! Now that we have both upstart and systemd, it'd nice if we can clarify which setup they should choose. According to Ubuntu, the systemd is only fully supported in Ubuntu 15.04 and later releases.

goetzk:
On 11/08/2015 01:24, Jenny Qian wrote:
> Great addition for adding systemd setup! Now that we have both upstart
> and systemd, it'd nice if we can clarify which setup they should choose.
> According to Ubuntu, the systemd is only fully supported in Ubuntu 15.04
> and later releases.

While 14.04 is the recommended platform so Upstart instructions should
remain for that. Perhaps a message to that effect will do the job.

PR #437, opened 2015-10-03
Reintroduce psql extension setup

goetzk:
This was first introduced in a commit by ed
::
    commit 9a98b4fc2e0a0c954d6bfa579b8e0f199f512b43
    Author: Ed Schipul <eschipul@tendenci.com>
    Date:   Wed Aug 26 20:21:05 2015 -050

but was unfortunately removed by mistake during my commit the next day - i'm
guessing i hadn't correctly updated my repo at the time.
::
    commit 621c5a1304380354adebfc3979b63510039f0813
    Author: Karl Goetz <karl@kgoetz.id.au>
    Date:   Thu Aug 27 18:39:51 2015 +1000

### DIFF
--- a/docs/source/installation/installation.txt
+++ b/docs/source/installation/installation.txt
@@ -168,6 +168,9 @@ Create the database and a new postgres user with the following commands (replace
     sudo -u postgres psql postgres -c "CREATE DATABASE $DB_NAME WITH OWNER $DB_USER;"
     sudo -u postgres psql postgres -c "GRANT ALL PRIVILEGES ON DATABASE $DB_NAME TO $DB_USER;"
     sudo -u postgres psql postgres -d $DB_NAME -c "CREATE EXTENSION postgis;"
+    sudo -u postgres psql postgres -d $DB_NAME -c "CREATE EXTENSION postgis_topology;"
+    sudo -u postgres psql postgres -d $DB_NAME -c "CREATE EXTENSION fuzzystrmatch;"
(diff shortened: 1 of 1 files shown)


jennyq:
Currently tendenci is not using those three extensions. Will need to check with @eschipul.

eschipul:
@jennyq - I did not make a new commit, but on my dev ubuntu server box I did need those extra GIS extensions to use PostGIS with Leaflet.js. I'll be doing another install from scratch next week to go through the instructions so I'll double check then.

This GeoDjango app does state that those extensions are needed as well
https://gist.github.com/hallahan/60c956987bde4a02b6be

    -- Enable PostGIS (includes raster)
     CREATE EXTENSION postgis;
     -- Enable Topology 
     CREATE EXTENSION postgis_topology;
     -- fuzzy matching needed for Tiger
     CREATE EXTENSION fuzzystrmatch;
     -- Enable US Tiger Geocoder
     CREATE EXTENSION postgis_tiger_geocoder;

the goal is to have the database ready so an app could plot the addresses on a map to display membership over time for example. GeoDjango has functionality similar to what I need for membership data mapping - hence the inclusion of the extra extensions. I don't want to have to go back and add them to every database later and I don't see much of a downside. 

And because I have already installed them I can't remember what error came up causing me to add them.

It's also possible that I'm crazy and was just getting happy installing stuff..... just can't remember the specifics of why I added those.

eschipul:
To clarify, I'm closing this pull because it is on the 7.1 branch. We can continue discussing here but the pull request in the short term should be on the master branch. We'll get to nightly dev builds soon, but for now, even if we're playing with fire a bit, we need to go straight to master.

PR #438, opened 2015-10-03
Minor update docs

goetzk:
Hi guys,
Not sure if this is a useful thing to have but here are some minor update commands i've picked up as i've experimented with tendenci. If something like this is desirable i'll fix the tiem typo (oops) and remove the footer before re requesting a pull.

PR #442, opened 2015-10-14
Minor update docs (for master)

goetzk:
This follows on from #438 , adding commands to be run for minor updates.

Still to be decided are the following:
* Best way to handle minor version updates
 - pip install -r requirements/tendenci.txt --upgrade.
 - a versioned upgrade using pip?
* Should it be updated to talk about tendenci 7.1 right away


eschipul:
Agreed. Great common sense input. We can be too geeky sometimes. Thank you!

PR #467, opened 2015-11-23
Run tendenci as www-data

### LINKED ISSUE #461: Drop root privileges in Upstart job
Hi,
I realised while doing some work today that the upstart job doesn't set the UID/GUID that it runs under in the configuration file.
I assume that means it runs as root?

If thats the case I suggest setuid/gid should be added to the upstart script in the documentation
setuid www-data
setgid www-data

If you agree I'll organise a branch for it.

goetzk:
This fixes #461 by running tendenci as www-data. I've done a quick test on an ubuntu system and it works so far (running as www-data via systemd appears to be fine too).


jennyq:
Thanks!

PR #485, opened 2016-02-21
Change tax calculation

goetzk:
Hi,
Rather than forcing the user to put in some weird not-quite-percentage number I
suggest putting the division by 100 in code rather than in the users head.

I'm not sure if this will require changes to existing systems or custom migrations - I'm hoping you can clarify that for me.

### DIFF
--- a/tendenci/apps/memberships/models.py
+++ b/tendenci/apps/memberships/models.py
@@ -378,7 +378,7 @@ class MembershipSet(models.Model):
         tax = 0
         if app and app.include_tax:
             invoice.tax_rate = app.tax_rate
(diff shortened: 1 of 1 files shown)


eschipul:
Closing this. We can probably handle this with a clarified literal below entry. I can see both sides and the software SHOULD be more user friendly, however given the gravity of financial calculations this isn't a change we can make lightly as every calculation, including ones clients have built in the wild that may not have been submitted as a pull request, would be impacted. And I believe there are several places where tax rates are entered which means for consistency we'd have to scrub those too.

Thus for expediency I'm going to close this one out as much as i appreciate the request.

PR #494, opened 2016-05-02
Remove odd character at end of file

jamesoflol:
Two newlines at end of file. The second last one seemed to be a bizarre ASCII character, which was interpreting into &#8203; and a bunch of a newlines, and thus style issues. I've removed this.


jennyq:
Thanks!

PR #504, opened 2016-07-07
* Refactored in-line choices values

mikecee:
Stylistic improvement; in-line `choices` arguments moved into class variables.


jennyq:
Nice! Thanks!

PR #532, opened 2016-11-19
Fix photo download filenames

PaulSD:
When downloading photos (http://example.com/photos/download/1/640x480/),
the filename in the Content-Disposition header included the full path to
the file on the server, but with slashes replaced with dashes.  This was
both an annoyance (as the user is not likely to want the path included
in the filename), and potentially a security issue (as it may provide an
attacker with useful information about the Tendenci installation on the
server).

This commit ensures that the path is removed from the filename before
adding it to the Content-Disposition header.


jennyq:
Great catch! Thank you!

PR #533, opened 2016-11-20
Replace SWFUpload with Fine Uploader

PaulSD:
Adobe Flash is not supported on most mobile platforms, it is no longer supported in Chrome/Chromium, and Firefox is planning to drop support for it in 2017.  Because of this, many users cannot upload files to Tendenci.  In addition, SWFUpload has been abandoned for many years and has well known security issues.

This series of commits replaces SWFUpload in Tendenci with Fine Uploader.  Fine Uploader is a modern, actively maintained, pure HTML/JS uploader.


jennyq:
Thank you very much for getting this in place for Photos and Theme Editor! We wanted to switch away from SWFUpload long time ago, and did change some but not all, for example, uploader in wysiwyg is replaced with the jQuery-File-Upload.

PaulSD:
The purpose of the script is really just to document the process for updating the FineUploader files that are included in Tendenci core, in case a Tendenci developer needs to update the version of FineUploader that is included in Tendenci core.  It is not meant to be run for each individual site, and it is not strictly needed if you never plan to update FineUploader.

jennyq:
Cool. Thanks!

Issue #523, opened 2016-10-22
App menus not visible to unauthenticated users


aprildh31:
Unauthenticated users no longer see any app menus.

For example, the calendar should have an "Events" menu that lets the user download an "Events ICS" file, or switch from the "Monthly Calendar" to the "Weekly Calendar" or "Events List", as shown here in Tendenci 5: http://metro.org/events/month/

As of Tendenci 6, this "Events" menu is part of the top menu for authenticated users, but cannot be accessed by unauthenticated users.

It looks like the old menus were removed by the following two commits:
https://github.com/tendenci/tendenci/commit/4296f627ef73b0529623b49df9aa6e406f3cda15
https://github.com/tendenci/tendenci/commit/24b1baf11c82003969fbbe100628914aea8fbf7b

It seems that @maxbixly had assumed that the new top menu replaced all of the functionality of the old menus, when in fact it does not replace the functionality of the old menus for unauthenticated users.

Any suggestions for how this can be corrected so that unauthenticated users can access these features? Is there any reason not to add the old menus back in, but only render them for unauthenticated users?

eschipul:
Three is a site setting to hide drop-downs (the white legacy ones) if that is the issue?

eschipul:
I believe this is resolved. Open a new issue if not please. also be sure you are on the 7.3 or whatever the latest version is.

PR #535, opened 2016-11-21
Restore app menus for unauthenticated users

PaulSD:
See #523.

These commits restore the old nav menus, but display them only for unauthenticated users.


eschipul:
@PaulSD - the tabs are redundant with what should now be in a given site's primary nav. The problem with showing them conditionally when there is no authenticated user is several fold.

1. Redundant security checks - there are already too many and we need a focus on greater speed once everyone gets done with the T5 to T6 to T7 process.
2. Tabs pushes content off/down the page on increasingly smaller viewports
3. Maintenance across devices with css without significant benefit

The original reason for the tabs was usability (2001, 2002 - html / tables), cross-browsers (IE4/Netscape battles), and lastly to increase internal links focused on high content areas (like articles search - a rich content page) for google bot.

eschipul:
Have you thought about doing a similar conditional within an include in your theme if you really want the tabs back? 

Leaving the pull open for further comments but as written it won't work.

PaulSD:
I agree this is a bad solution.  My thinking here was "Something is clearly broken, but I don't have the foggiest idea how to fix this properly, and we never got a response to the issue we opened about it, so until we can figure out something better I'll just revert the commit that broke it, and maybe submitting a pull request for that will elicit a response."  I guess that worked as intended. ;)

When I initially reverted this, I was specifically interested in restoring the old nav in the Events app, but I figured that there might be similar navigation issues in other apps, so I restored the links in all of the apps so that I could find those other issues.  I have since gone through all of the apps more carefully, and I believe there are only two places where the missing old nav is an issue: the Events app and the Locations app.

The issue with the Locations app is that there are two searches: The search box on the locations list (/locations/), and the nearest search box that is on its own page (/locations/nearest/).  Without the old nav, there is no way for a user to move between those searches without going through the site's primary nav, which is totally unintuitive.  I think a simple solution for this is to add an option to the primary (location list) search box so that you can perform either a "normal" or a "nearest" search from the same page.

One problem with the Events app is that it has several different views (List/Monthly/Weekly, plus the Print view and the ICS download) which users are likely to want to be able to switch between.  However, without the old nav, there are no links within the page itself to switch between those views, and the site's primary nav is not the appropriate place for those links.

There are two major problems with putting those Events app links in the primary nav:
1. The primary nav should be clean and simple so that new users can find the general area of the site that contains the information they are looking for with as little reading/searching/decision making as possible.  Adding 4 additional items to the nav for the additional Events views will clutter up the primary nav significantly and may confuse users ("Why are there five links to the same Events list?").  It might be reasonable to nest those additional links as a 3rd level items under the 2nd level nav link for Events (which itself could go directly to whatever view we decide should be the Events app's primary view).  However, 3rd level navs are generally discouraged in modern web pages, and they are no longer supported in Bootstrap 3 (https://github.com/twbs/bootstrap/pull/6342#issuecomment-11594010).
2. If the user arrives at the Events app without going through the primary nav, then there is no good way for them to discover that they can switch between views.  For example, if someone clicks a "Click here to see our upcoming events" link in some page or email, then they may not have seen the additional links in the primary nav, and there will be no intuitive way for them to discover that there are alternate views which might be useful to them.

A second problem with the Events app is that on the listing page for a specific event, without the old nav, there are no links to download the ICS for that event or to go to the "View Recurring Event List" page associated with the event.  It is not possible to dynamically add links for these event-specific pages to the primary nav, so without the old nav there is no way to access those features.

It seems to me that these links need to be added somewhere within the Events pages themselves.  Perhaps the side-bar is the appropriate place for them?

eschipul:
I see your point and I agree with the functional requirements. In fact a regular user SHOULD be able to add an event, retain ownership of it, only having it go live after a super_user approves it. This would then allow the regular user (as owner of the event) to continue to have full permissions to that event including running the roster report, exports, and invoices related to that event.

Business case: why should an executive director have to add events for everyone? Why can't I run a report on MY EVENT even if I'm not a super_user? Tendenci is supposed to be about empowerment with security as well as checks-and-balances. But not to the exclusion of functionality and distributing responsibility. Overly controlling associations aren't the target audience I was going after, while still respecting the need for centralized control applied judiciously.

Users adding events and other functionality as owners of an object - Tendenci was built that way in 2001. Lost in translation in 2009 rewrite. 

The only good news is we can learn from the original design pattern. There were three methods that handled the sub-navigation of a module. A non-logged in user would see a page like this when they went to the index of any folder (urls.py now) such as http://web.archive.org/web/20100303013659/http://www.schipul.com/en/calendarevents/

While clearly having html mixed with code is not acceptable now, it could be achieved with template includes, the methods were:

calendarevent.showtabs(activeuser, *kwargs) - white tabs for high seo content pages
calendarevent.showh(activeuser, *kwargs) - horizontal text nav at the top below tabs
calendarevent.iconrow(activeuser, *kwargs) - icon row to the right at top for fast navigation

Example event view
http://web.archive.org/web/20100130203547/http://www.schipul.com/en/cev/493

The question is how best to return that functionality. It can't be in the black bar at the top as that is for logged in only. If it could be done FAST with the django has_perm then the old method of rendering a page with fewer than 7 queries would work. But the load is too great for a whole other set of technical reasons.

I'd love to see this come back to life in issues for further input. Mainly because your request IS CLEARLY IN THE SPIRIT OF WHAT TENDENCI SHOULD BE!

Alas for now I'm going to close this pull request until we can com up with a middle ground.

Ed

PR #538, opened 2016-11-27
Add a list_photo_sets django template tag

PaulSD:
The existing list_photos django template tag makes it easy to get a list of photos in a custom template, but there is currently no good way to get a list of photo sets in a custom template.  This commit adds a list_photo_sets django template tag.


jennyq:
Nice! Thanks!

PR #540, opened 2016-12-19
Fix search indexes

PaulSD:
It looks like the updating of most search indexes has been broken since 7.2.0 was released.  You can still search old entries that were added to the search indexes before upgrading to 7.2.0, but any new content that is added to a Tendenci site after upgrading to 7.2.0 will not show up in the search results.  Indexing of events and forms_builder items may have been broken since before 7.2.0 (I'm not sure exactly when those broke).

These commits fix the indexes so they are now updated as new content is added.

However, it doesn't look like there is any way to re-index existing content.  (tendenci/apps/search/management/commands/process_unindexed.py will index queued items, but these search index problems prevented new items from even being added to the search index queue, so that doesn't help.)  Someone might want to write a script to re-index all existing content to clean this up on existing sites.

### DIFF
--- a/tendenci/apps/events/utils.py
+++ b/tendenci/apps/events/utils.py
@@ -10,6 +10,7 @@ from dateutil.rrule import *
 from decimal import Decimal
 
 from django.contrib.auth.models import User
+from django.core.exceptions import AppRegistryNotReady
 from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
 from django.db import connection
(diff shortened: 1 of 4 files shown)


jennyq:
Great catches!  And I love the detailed description of your commits! Nice job! 👍

jennyq:
Regarding re-index existing content, it is not necessary because every add, edit or delete will insert an entry to the UnindexedItem. The `process_unindexed` command will call `update_index` command with the `--age` option set to the number of hours back to the oldest items.  

If you do want to rebuild index, you can run `rebuild_index` command http://django-haystack.readthedocs.io/en/v2.5.0/management_commands.html#rebuild-index

PR #545, opened 2017-01-03
filenames in Content-Disposition headers should be quoted

PaulSD:
Without this, filenames with spaces and/or other special characters get mangled.


eschipul:
Thanks @PaulSD !

PR #546, opened 2017-01-09
Fix handling of hidden folders in theme_editor

PaulSD:
The previous code would throw an exception if a non-hidden folder was
nested inside a hidden folder.  This is because os.walk() traverses into
hidden folders, and the code searches for the parent of each non-hidden
folder in files_folders, but it doesn't add hidden folders to
files_folders, and it doesn't handle the case where the parent is
not found in files_folders.  So a non-hidden folder in a hidden folder
causes the parent search to fail.


PaulSD:
That is exactly what this commit does.  It still excludes hidden files and folders, but it now also excludes folders nested within hidden folders.

jennyq:
Excellent! Thanks!

PR #556, opened 2017-02-27
Fix view permissions on objects with 1 digit primary keys

PaulSD:
If search indexes are available, has_perm() in apps/perms/backend.py
calls can_view() in apps/perms/utils.py to determine whether a user has
view permissions on an object.  can_view() queries the index for
documents matching (SQ(primary_key=obj.pk) & (q_groups | q_users)) and
returns True if any matching documents are found.

Since the primary_key index field is defined as a CharField, Whoosh
applies StopFilter to the query for that field.  If the query value is
less than 2 characters long (if the object's primary key happens to be
<= 9) then StopFilter removes that query from the query set, reducing
the query to just (q_groups | q_users).

That causes documents to be returned (and can_view() to return True) if
the user has view permissions on any objects in the index.

This commit fixes the issue by changing primary_key to an IntegerField
such that StopFilter will not be applied to queries on that field.

In addition, a sanity check is added to can_view() to help ensure that
the result set returned by the index makes sense.

### DIFF
--- a/tendenci/apps/perms/backend.py
+++ b/tendenci/apps/perms/backend.py
@@ -169,6 +169,8 @@ class ObjectPermBackend(object):
                 index = site.get_index(obj.__class__)
                 if can_view(user, obj):
                     return True
+            except AssertionError:
+                raise
             except:
                 pass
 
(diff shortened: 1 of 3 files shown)


jennyq:
Great catch!

I found the result is opposite - if the object's primary key happens to be <= 9, it just returns False regardless.

In the meantime, for existing sites, it is needed to run `update_index` command to update whoosh index once this fix is deployed.

Issue #555, opened 2017-02-25
Unable to Delete Corporate Members Types or Applications


FourthEstate:
Version: 7.2.48

I am not able to delete Corporate Member Types or Corporate Member Applications.  

I've looked at debugging and nothing jumps out and no error messages are being logged. 

I am however able to ADD and MODIFY Corporate Member Types and Applications

I double checked demo.tendenci.com and this behavior persists in the demo as well. I am able to Corporate Member Types or Corporate Member Applications but not able to delete them in in the demo either.

FourthEstate:
I double checked demo.tendenci.com and this behavior persists in your demo as well.  I am able to Corporate Member Types or Corporate Member Applications but not able to delete them in in the demo either.

PR #558, opened 2017-02-27
Don't show soft-deleted objects

vladkorkach:
Solution for #555 
I've added a filter which doesn't show data marked as deleted.


jennyq:
Thank you!

FourthEstate:
vladkorkach  contact me for some work.

FourthEstate:
Note that this introduces a new issue where additional corporate membership types can't be added if a "hidden" category has same name.  

Delete should really be a DELETE, not just a hide

screen shot 2017-03-12 at 3 49 17 pm

jennyq:
@FourthEstate, the delete is set to soft-delete for a reason so that if someone accidentally deleted an item and wanted it back, it'd be easy to pull back. However there **IS** a way to physically delete the soft-deleted items - just run the command `delete_soft_deleted_items`. You can add this command to the cron jobs to run nightly if needed.

Issue #562, opened 2017-03-21
Bug: Corporate Membership Form Doesn't Obey Corp Member Type Grouping


FourthEstate:
Despite being grouped by membership type the membership form ignores this and displays ALL membership types.  

GROUPING
![screen shot 2017-03-21 at 10 40 59 am](https://cloud.githubusercontent.com/assets/23258672/24152755/eaff13de-0e22-11e7-907f-2b76c588adcf.png)

OUTPUT
![screen shot 2017-03-21 at 10 43 03 am](https://cloud.githubusercontent.com/assets/23258672/24152846/36107052-0e23-11e7-85b9-1b7511712d14.png)

eschipul:
Closing issue and assigned pull request https://github.com/tendenci/tendenci/pull/564 for testing. Thank you for being part of the community!

Issue #563, opened 2017-03-21
Bug: Corporate Membership Applications Don't Honor Payment Method Selections


FourthEstate:
There is another issue where the Corporate Membership application form doesn't honor the selections made on creation or editing.  The result is that ALL available payment options are shown, not just the ones that were selected.

eschipul:
Thank you for the fix and the pull request https://github.com/tendenci/tendenci/pull/564 . Assigned for testing and will push out with next version bump.

PR #564, opened 2017-03-23
Fixed Bug #562 & #563 - Fixed possible copy & paste error that caused…

casbeebc:
~~Fixed possible copy & paste error that caused issues with the corporate membership form fields being incorrect. Issues #562 & #563~~

Seems to be  an issue with iterating though a field's choices not via the field widget's choices. So it will display all choices and not the filtered choices. Fixes issues #562 & #563


FourthEstate:
This fixed the issues in #562 and #563  but introduced some form styling issues

jennyq:
Thank you!

PR #569, opened 2017-04-24
Allow multiple CoreMembershipApp and Clone method 

### LINKED ISSUE #560: BUG:  Corporate membership only creates one membership application link
We are seeing an issue where only one membership form URL link is being created despite multiple forms and corporate membership types being setup.  

The Corporate membership module only allows us to create one membership application.  However the ability to create individual membership applications based on the type of Corporate Member would be immensely useful.

For example we have News Organization members but also other types of organizations such as associations, colleges/universities etc that we want to have unique memberships types and applications for. @

sonus21:
In this change, multiple things have been done
1. In current design, Tendenci can have only one CoreMembershipApp this is all due to current_app method in CoreMembershipApp, this change has removed this restriction and fixes #560 
2. Add Clone method for Corporate Application in admin
3. Allow Multiple CorpMembershipApp to reuse same MembershipApp

### DIFF
--- a/tendenci/apps/corporate_memberships/admin.py
+++ b/tendenci/apps/corporate_memberships/admin.py
@@ -50,11 +50,11 @@ class CorporateMembershipTypeAdmin(admin.ModelAdmin):
             '//ajax.googleapis.com/ajax/libs/jqueryui/1.11.0/jquery-ui.min.js',
             '%sjs/admin/admin-list-reorder.js' % settings.STATIC_URL,
         )
-        
+
     def get_queryset(self, request):
         qs = super(CorporateMembershipTypeAdmin, self).get_queryset(request)
         # filter out soft-deleted items
-        return qs.filter(status=True)        
+        return qs.filter(status=True)
 
     def save_model(self, request, object, form, change):
         instance = form.save(commit=False)
@@ -86,6 +86,13 @@ class CorpMembershipAppFieldAdmin(admin.TabularInline):
     template = "corporate_memberships/admin/corpmembershipapp/tabular.html"
 
 
+def clone_apps(model_admin, request, queryset):
+    for form in queryset:
+        form.clone()
+
+clone_apps.short_description = 'Clone selected forms'
+
+
 class CorpMembershipAppAdmin(admin.ModelAdmin):
     inlines = (CorpMembershipAppFieldAdmin, )
     prepopulated_fields = {'slug': ['name']}
@@ -111,6 +118,7 @@ class CorpMembershipAppAdmin(admin.ModelAdmin):
     )
 
     form = CorpMembershipAppForm
+    actions = (clone_apps,)
 
     class Media:
         js = (
--- a/tendenci/apps/corporate_memberships/migrations/0001_initial.py
+++ b/tendenci/apps/corporate_memberships/migrations/0001_initial.py
@@ -377,7 +377,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='corpmembershipapp',
             name='memb_app',
-            field=models.OneToOneField(related_name='corp_app', default=1, to='memberships.MembershipApp', help_text='App for individual memberships.', verbose_name='Membership Application'),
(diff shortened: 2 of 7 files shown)


jennyq:
Thank you @sonus21 ! 

With multiple corp. apps, it does increase the degree of complexity for both corporate memberships and memberships.  Below are the two areas I can think of that are also affected by this change:

1. On individual membership join/renewal 

When individuals join/renew memberships under a corporate membership, they are authenticated based on the method specified in the corporate membership application. Previously, there is only one corp. app, so we just retrieve the info from current_app. Now we'll need to know the exact app this corporate membership tied to.

2. On corporate membership edit, which corp. app can be pulled to render? If a corp. membership is joined with corp. app A, you don't want to render with corp. app B, right?

So, each corp. membership needs to remember which corp. app it used. I'd suggest to add a ForeignKey field to reference CorpMembershipApp on CorpMembership. Thoughts?

sonus21:
Hi Jenny,
Thanks for the review. Your both points are valid, we should use a ForeignKey to CorpMembershipApp in CorpMembership, but I don't see any code which uses to current_app in CorpMembership,  but in some places, we have used in views. 

 Migration was just a quick hack, given that's only for Django internals since in DB Django doesn't make any update.

My thought would be to remove the concept of current_app from corporate_membership which will clean up the code.  I was thinking same during development, but that would be a big change, many templates and functions have to be changed.

If you think, I can take this up and will follow up. I'm not much aware of Tendenci coding style, still will follow up PEP8.

jennyq:
@sonus21,

Since the `current_app` will no longer be relevant, you can remove it from corporate_membership. Then update every place it is used. Yes, It would be a big change!

Regarding the migration from `OneToOneField` to `ForeignKey`, postgreSQL does add the unique constraint to the `OneToOneField`. So it's NOT just for Django internals. 

As for the coding style, yes, please follow PEP 8 and django's coding style if applicable: https://docs.djangoproject.com/en/1.8/internals/contributing/writing-code/coding-style/

Thank you very much for your contribution. We really appreciate it!

eschipul:
@sonus21 

Thank you for getting involved! 

I have some concerns and we all need to understand this is a HUGE change. Probably worth of a branch for a few months as I can see the possibility of some significant issues.

I'm not against it in principal, but I question the proposed removal of "current_app". (I backed down on this type of "simplification" with not having a user object instantiated on every page view in the jump from T4 to T5. Obviously I regretted this when on 1.4 (I think it was 1.4?) django fixed the framework and the anonymous user object was added. We then had to add it back. But lost all utility given we programmed to the previous standard. Uuuugh. 

Other possible scenarios on membership_app and why I view this proposed removal as a complication and not a simplification:

1) person A qualifies for membership under CHARITY1 and also under BIG-CORP. Publicly she would rather be associated with BIG-CORP for personal reasons (again given they qualify for both)

2) person B qualifies for BIG-CORP but big-corp limits qualifying corp membership at that sponsorship level to 5 individual members. person B might be an intern and NOT want to use one of the 5 slots that might prevent the CEO from claiming her slot out of the 5.

3) The django sites implementation. If a national organization like aama.org wants to use Sites for local chapters, allowing local admins to be super users at their level (different URL, same DB) then again we will be adding back the previous functionality of entities as pointers for multi site data publishing and permissions. In a tree of five sites (national - four chapters) how does one associate the individual membership with the correct app?

I will defer to @jennyq. But my gut is just telling me this will come back to bite the project. I prefer explicit references, even if they seem redundant and pedantic, it is better future proofing. 

We must implement Django Sites and tie entities into the pointer/tree for financial tracking. If that is a given, and it is, it seems like yet another reason to be more explicit. I can think of many other scenarios. Can y'all take the above into consideration and tell me if I'm not understanding it, or if the new information above changes your thoughts? 

Thank you again regardless. And @jennyq I will defer to your wisdom on this one.

jennyq:
This PR is no longer compatible with the latest version of Tendenci. @sonus21, if you're still working on it, please submit a new PR. Thank you!

PR #573, opened 2017-05-19
Fix issue #560: Corporate membership only creates one membership application link

LuisLlave:
The proposed changes try to resolve issue [#560](https://github.com/tendenci/tendenci/issues/560).

I didn't changes in model because the issue are related to show urls for Corporate Membership Applications already created, and make working those urls.

### DIFF
--- a/tendenci/apps/corporate_memberships/models.py
+++ b/tendenci/apps/corporate_memberships/models.py
@@ -1324,6 +1324,10 @@ class CorpMembershipApp(TendenciBaseModel):
 
         return current_app and current_app.id == self.id
 
+    def is_active(self):
+        """Check if status application is 'active'."""
+        return self.status and self.status_detail == 'active'
+
     def save(self, *args, **kwargs):
         if not self.id:
             self.guid = str(uuid.uuid1())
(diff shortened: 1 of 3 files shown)


jennyq:
Hi @LuisLlave ,

To resolve issue #560, there are more things needed to be addressed than the links for the add pages. For example, when rendering the edit and detail pages, it needs to know what app was initially used. Please take a look at how membership module deals with multi applications. 

In addition, when individuals join/renew memberships under a corporate membership, they are authenticated based on the method specified in the corporate membership application. That should also be take account into when enabling multi corp applications.

Thanks!

LuisLlave:
@jennyq Hi, reviewing code, I think that I need to change all views and methods using **current_app** method, and pass app id to them in order to get data from correct app; then, current_app will be unnecessary and can be deleted.

Is that approach correct? You will accept that proposed change?

jennyq:
Hi @LuisLlave, yes, I think that approach is the only way to make it work. You can follow the same pattern of how memberships handles the multi-apps.

BenSturmfels:
@jennyq I believe you're safe to close this merge request as it is a less developed version of #569.

PR #574, opened 2017-05-26
Also set status_detail when setting TendenciBaseModel.status.

BenSturmfels:
When setting `status` to False, `status_detail` should also be set to `inactive` for consistency.

### DIFF
--- a/tendenci/apps/perms/models.py
+++ b/tendenci/apps/perms/models.py
@@ -125,6 +125,7 @@ class TendenciBaseModel(models.Model):
         # set status to False and then save(). We do NOT
         # actually delete anything from the database.
         self.status = False
+        self.status_detail = 'inactive'
 
         # Making slugs unique by appending pk
         # This prevents Integrity errors if


jennyq:
@BenSturmfels I see your point. But sometimes people accidentally deleted an item and wanted it back. So it's better to keep the original status detail in this case.

eschipul:
@BenSturmfels to your point, in some ways I agree but in the bigger picture this does diverge from the model as designed.

An example would be many people accidentally delete their nav objects and then panic. Setting it back to the boolean "1" from "0" is a quick fix. With navs of course that means super-users. The original intent, and something we have strayed from, is a default of allowing regular users to add content like events and then have it approved (strictly 1 level of approval) by the administrator for a given module.)

Event logs and the object itself should be a snapshot in time of the state of the object at the time of delete. Hence changing any other properties of the object at the time of "soft delete" would be inconsistent with the intent.

Before closing, I'd like to ask for your input in case I haven't explained the intent, or more importantly to ask if the intent itself is flawed in which case I'm very open to further conversation on the topic. Thoughts?

BenSturmfels:
Thanks @jennyq and @eschipul. The background on this is that I was troubleshooting an unrelated membership issue and saw that a MembershipType record was marked `status = false` and `status_detail = "active"`. I naively assumed that these two fields should be in lock-step, but it sounds like there's more to it. I'll close this one.

BenSturmfels:
(Edited: Meant to say MembershipDefault rather than MembershipType above.)

PR #577, opened 2017-05-30
Add NoticeType "membership_notice_email", remove bare except.

### LINKED ISSUE #576: Membership notices not sent due to missing NoticeType "membership_notice_email"
When I run `python manage.py send_membership_notices`, I've found that the notices are not sent due to a missing `NoticeType "membership_notice_email"`. There's a bare `except` in [`notifications/send_emails.py`](https://github.com/tendenci/tendenci/blob/v7.2.77/tendenci/apps/notifications/models.py#L341l) that silently handles `NoticeType.DoesNotExist` errors, when these should probably be flagged.

I have a range of NoticeTypes installed via `memberships.signals.create_notice_types`, but not "membership_notice_type". The only mention of this type is in `send_membership_notices`. Could it just have been missed?

Will attach a patch shortly.

BenSturmfels:
…eTypes.

The "membership_notice_email" NoticeType is referenced in
`apps.memberships.management.commands.send_membership_notices`, but not created
during migrations.

Additionally remove bare except and add some logging. Bare excepts are generally
a bad idea because they hide the cause of problems.

Fixes #576.

### DIFF
--- a/tendenci/apps/memberships/signals.py
+++ b/tendenci/apps/memberships/signals.py
@@ -17,7 +17,7 @@ def update_membs_app_id(sender, **kwargs):
     app_to_be_deleted = kwargs['instance']
     switch_memberships_app_id(app_to_be_deleted)
 
-       
+
 def switch_memberships_app_id(app_from):
     # each membership has an app_id associated.
     # since this app is to be deleted, we need to update memberships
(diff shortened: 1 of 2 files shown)


jennyq:
Great catch! THANK YOU!
I did a search and found why that notice was missed. It was **not** included in the list along with other types in the first place. Instead, it was added to the database migrations in the old T5 version eacd81a .

PR #579, opened 2017-06-05
Move "generated by Tendenci" email footer to a template include.

BenSturmfels:
Currently this email footer is defined in a number of places in both Python code
and Django templates. This change centralizes the footer into a single location
that can be overridden with templates.


BenSturmfels:
I've tested the `send_membership_notices` and renewal paths, which covers most of these changes, but have not tested the similar changes made to corporate memberships as they aren't used on my current project.

jennyq:
Very nice! This will make the footer easy to maintain. Thanks for eliminating the duplicates and making the code DRY!

PR #580, opened 2017-06-07
Fix raw HTML showing up in renewal alert.

BenSturmfels:
Issue was introduced by applying `safe` template filter in 79fd1e3120191aceae4d95954331f59b00e07d0f.


BenSturmfels:
Screenshot showing raw HTML in membership renewal alert before this patch is applied:

screen shot 2017-06-07 at 13 35 16

And after:

screen shot 2017-06-07 at 13 40 41

jennyq:
Thank you!

Issue #595, opened 2017-08-10
Add support for Python 3


BenSturmfels:
While adding support for Python 3 is quite a lot of work, it makes Tendenci a more attractive platform to build on and contribute to. The process of adding this support also tends to flush out hidden bugs related to with text encoding.

PaulSD:
Note that Django 2.0 will drop support for Python 2, so April 2020 (end of support for Django 1.11) is a hard deadline for getting this done.

PaulSD:
Fixed in #659

BenSturmfels:
@jennyq and @PaulSD: I'd be really interested to hear your thoughts on how you'll be approaching the the Python 3/Tendenci 8 upgrade with existing clients. Do envisage that you'll be switching existing clients to T8, or only new clients?

Given that the T8 branch requires a substantially modified project-template, it sounds like this is going to be relatively difficult upgrade path. It sounds like the easiest path may be to load the existing T7 database into a fresh T8 project-template and then port across any custom settings and apps.

Would you be interested in contributions to make T8 compatible with a T7 project template, so that the upgrade could be done incrementally?

@PaulSD thanks for the mountain of work you've done on all this!

PaulSD:
T8 can use an existing project template with only minor changes.  See "Step 4b" in the upgrade instructions: https://github.com/tendenci/tendenci/blob/tendenci8/docs/source/upgrade/upgrade-to-tendenci8.txt#L116  :)

BenSturmfels:
Thanks @PaulSD, that's perfect!

jennyq:
Thanks @PaulSD !

Not sure yet, but we'll try to script the upgrade process if possible.

PR #596, opened 2017-08-10
Apply Python Futures automated "futurize" script,  stage 1.

BenSturmfels:
This applies fixes that modernize Python 2 code without changing the effect of
the code. The changes include fixing print statements, replacing relative
imports, replacing ``has_key`` and using new-style exception handling.

These changes are in theory be safe to apply, but should be thoroughly tested
before being merged to master. I've done some light testing of the profile and
membership app.

For details, see:
http://python-future.org/futurize.html#forwards-conversion-stage1

Related to issue #595.


jennyq:
Thank you very much for starting on this! Making Tendenci Python 3 compatible is very important. Since the process will involve a lot of changes, let’s add commits to the branch “python3_compatible”. Once the process is completed and thoroughly tested, we’ll merge it to master.

PR #604, opened 2017-09-17
Fix subprocess venv, improperly disabled links, and improperly disabled registration forms

PaulSD:
See commit messages for details


jennyq:
Nice fixes! Thank you!

PR #609, opened 2017-10-17
SECURITY: Adjust gzip encoding to prevent BREACH attacks

PaulSD:
"BREACH" refers to a class of attacks that use on-demand gzip encoding to determine the content of an encrypted connection without actually decrypting it.  Due to this class of attacks, it is not safe to use gzip encoding for most encrypted dynamic content.

Tendenci is currently configured to gzip encode all content generated by Django, making it vulnerable to BREACH attacks.  In addition, the example nginx config in the Installation document is not configured to gzip encode most static content served by nginx, which likely means that most Tendenci installs are not gzip encoding the static content that is actually safe to gzip encode.

These commits fix these two problems, disabling gzip encoding for dynamic content and enabling gzip encoding for static content served by nginx in the example nginx config.

### DIFF
--- a/docs/source/installation/installation.txt
+++ b/docs/source/installation/installation.txt
@@ -537,6 +537,9 @@ First, make an nginx configuration file for the site. This will be created at ``
         charset   utf-8;
         keepalive_timeout  65;
         client_max_body_size  30M;
+        gzip_types text/css application/javascript text/javascript text/plain text/xml application/xml;
+        gzip_vary on;
+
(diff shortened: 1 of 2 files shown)


jennyq:
Thank you!

eschipul:
@PaulSD - anyone who takes security as serious as you do is a friend of everyone using Tendenci. THANK YOU! 

And thank you for the merge @jennyq - YOU ROCK!

PR #612, opened 2017-10-30
Lots of fixes

PaulSD:
Sorry for the large dump.  I have quite a few changes that should have been pushed long ago, but I never got around to refining / organizing them enough to push them.  However, the delta between my code base and your code base is getting unwieldy, so I'm trying to clean some of this up now.

### DIFF
--- a/docs/source/installation/installation.txt
+++ b/docs/source/installation/installation.txt
@@ -52,7 +52,7 @@ You'll need those packages installed on Ubuntu.
 ::
 
     sudo apt-get install build-essential python-dev libevent-dev libpq-dev libjpeg8 libjpeg-dev libfreetype6 libfreetype6-dev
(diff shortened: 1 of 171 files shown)


PaulSD:
> Maybe the truncatechars filter is a better alternative here.
Your review comments show up on GitHub as normal comments - they aren't associated with specific commits or line numbers - so it took me a bit to figure out what you were referring to...

Regardless, I've updated the commits to fix both of your comments.

jennyq:
Sorry about that. I didn't realize github won't associate the comment with the commit being reviewed. I should have added the comment to the specific line of code instead of hitting the "Review changes" on a commit.

jennyq:
Great job! Thank you so much for so many fixes!!

PR #614, opened 2017-11-02
SECURITY: Prevent fraudulent simultaneous reuse of PayPal transactions

PaulSD:
See the comments at https://github.com/tendenci/tendenci/commit/dba60d3ab7991e8a7ff2984d7bf74dd66b028794#diff-dd6e5f6babec8c30bf34f8fdd94c45fbR123 for details.

The second and third commits also prevent non-fraudulent duplicate processing of all payment requests.


tendenci/apps/payments/paypal/utils.py
```diff
@@ -119,8 +120,28 @@ def verify_no_fraud(response_d, payment):
...
+    # transaction.atomic() with select_for_update() ensures that
+    # payment.trans_id cannot be updated in the database by another process
+    # after this calls get_object_or_404() and before this calls payment.save().
+    transaction.commit()
+    with transaction.atomic():
+        payment = get_object_or_404(Payment.objects.select_for_update(), pk=payment.id)
+        if payment.trans_id != '':
+            if payment.trans_id != txn_id:
+                return False
+            payment.trans_id = txn_id
```
jennyq: At this point, it's unsure if this txn_id is legitimate. Should it really be saved? What if a valid later transaction has this txn_id?

PaulSD: Good point.  This code was designed to ensure that invalid requests are never accepted, and it was not designed to ensure that valid requests are always accepted.  However, we could move the duplicate txn_id check to the end so that it is the last check before writing the database record.

PaulSD:
I've revised these commits to handle some additional corner cases and to improve the comments in the code.

jennyq:
Way to go, @PaulSD !

Regarding the log_silent_post(), it's okay to disable it. It was basically the legacy code used primarily for troubleshooting. But with nginx logs, there is not too much value of it. And I agree the default_storage is not a good location for that.

PR #618, opened 2017-11-10
Fix manage.py hangs when memcache is enabled

PaulSD:
This is another permutation of 0f0de92c and ae9640b7 and 5314592e ...
django.setup() loads haystack, which imports a bunch of Tendenci models
and indirectly imports apps/base/utils.py.  apps/base/utils.py calls
get_theme_root() at import time, which calls get_setting(), which calls
memcache.get(), which hangs if called from within django.setup().

To fix the issue, apps/base/utils.py should call get_theme_root() at run
time rather than at import time.

### DIFF
--- a/tendenci/apps/base/fields.py
+++ b/tendenci/apps/base/fields.py
@@ -145,10 +145,10 @@ class EmailVerificationField(fields.MultiValueField):
 class CountrySelectField(fields.ChoiceField):
     def __init__(self, *args, **kwargs):
         super(CountrySelectField, self).__init__(*args, **kwargs)
-        
+
         initial_choices = ()
         exclude_list = []
-        
+
         initial_choices_keys = get_setting('site', 'global', 'countrylistinitialchoices')
         if initial_choices_keys and initial_choices_keys != u'[]':
             initial_choices =  ((name, name) for key, name in list(COUNTRIES) if key in initial_choices_keys)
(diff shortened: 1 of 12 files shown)


PaulSD:
Hmmm... This was the only deadlock I managed to reproduce, so I must be missing something ... but I can try to test some more.

If you can get stack traces for the deadlocks you hit, I can help clean these up even if I can't reproduce them.  Easiest way to get a stack trace is to add a syntax error just before the cache.get() call so that it crashes the process immediately before it would otherwise deadlock.

jennyq:
Not sure why you cannot reproduce it. Here is the list of lines that contribute to the issue.  I haven't dig deep yet. Much appreciated if you can clean it up.

        tendenci/apps/base/utils.py 566 get_template_list     theme = get_setting('module', 'theme_editor', 'theme')

        tendenci/apps/forms_builder/forms/admin.py 73 FormAdmin     if not get_setting('module', 'recurring_payments', 'enabled'):

        tendenci/apps/files/utils.py 517 get_max_file_upload_size     global_max_upload_size = (get_setting('site', 'global', 'maxfilesize') or

        tendenci/apps/files/utils.py 520 get_max_file_upload_size         return int(get_setting('module', 'files', 'maxfilesize') or global_max_upload_size)

        tendenci/apps/base/fields.py 152 __init__         initial_choices_keys = get_setting('site', 'global', 'countrylistinitialchoices')

        tendenci/apps/base/utils.py 566 get_template_list     theme = get_setting('module', 'theme_editor', 'theme')

        tendenci/apps/corporate_memberships/settings.py 14 <module> use_search_index = get_setting('site', 'global', 'searchindex')

        tendenci/apps/corporate_memberships/admin.py 36 CorporateMembershipTypeAdmin     if get_setting('module', 'corporate_memberships', 'usefreepass'):

PaulSD:
I started cleaning those up, but some of those get_settings() calls are hard to avoid at import time without making some pretty large changes...

So, new approach ... let's just have get_settings() skip the cache.get() call if it is called from within django.setup().

jennyq:
Fabulous! THANK YOU!

PR #620, opened 2017-11-15
Fix some flake8 warnings

PaulSD:
No functional changes - simply replacing deprecated/discouraged syntax with recommended equivalent syntax.


jennyq:
Nice! Thanks!

PR #630, opened 2017-12-28
Specific Django version in installation instructions

hammygoonan:
Updating the installation documentation to be more specific about which version of Django to use.


hammygoonan:
Fair enough. To be honest, I think that rather than moving away from dj-pagination, the best solution would be to make the app work with Python3/Django2 but I realise that’s a much bigger concern and unrelated to this issue.

jennyq:
I totally agree.

PR #643, opened 2018-02-14
Defer executing DB queries

goetzk:
When importing memberships.reports or model_report.utils the database queries
are run immediately.
This causes problems when run through pytest-django (and possibly other tools)
which suppress unrequested DB access as a matter of course

I've opted to defer the queries here as I saw the same method used elseware in
those files and thought it was likely to be an acceptable solution.

I enquired with pytest-django (#573) if they had some official documentation they could point me to but unfortunately no response has been forthcoming.


jennyq:
Thank you! I've updated a bit to avoid the repeated queries.

goetzk:
Thanks!

PR #645, opened 2018-02-18
Mention recurring payments

goetzk:
This adds mention of recurring payments to the 'readthedocs' based
documentation and links to the existing pages where possible.

I was hoping to add more - such as how Tendenci defaults to monthly billing - but I haven't managed to figure out how to change those settings yet.


jennyq:
Nice! Thank you!

PR #647, opened 2018-02-24
Amend URL to help file

goetzk:
While the T5 setup page redirects, its cleaner to point directly to the correct
page - pleasingly the same page linked to in #645.


jennyq:
Nice catch! Thanks!

PR #650, opened 2018-02-28
Update requirements.txt

surajprasad:
BeautifulSoup==3.2.1 seems to crash the install with python3.6 in venv


PaulSD:
Simply removing the dependency on BeautifulSoup breaks code that hasn't been updated to support BeautifulSoup4 yet.  Fixed in #659

jennyq:
Thanks!

PR #652, opened 2018-02-28
Expand recurring payments documentation

goetzk:
When I submitted PR #645 I was still missing some information when it came to making use of the recurring payments.

Most of that missing information is now included here, I only need to figure out how to change the length of a member(ships)
payment cycle.


jennyq:
Thanks!

Issue #613, opened 2017-10-30
Migrate from Django 1.8 to 1.11


PaulSD:
Support for Django 1.8 will end in April 2018.

AdamBark:
I'm not really sure what the best way to communicate with the project is but this seems relevant. I started a branch to do this migration and have Django 1.9 working for migrations and it passes 'test' https://github.com/AdamBark/tendenci/tree/django-upgrade
I'm not sure everything I've done so far is exactly right but hopefully some of it will be useful. I've started working on 1.10 and I'll comment again when that's working and pushed to github.

eschipul:
Thank you again - we're sometimes get caught up in client work when we want to be working with the open source community more. There is of course no difference in the code - the open source version IS the exact same version our hosted clients use. So thank you again.

As for a major change like the move to 1.10, i believe it definitely warrants a new branch. I'm a big proponent of commenting code, moving fast, but we've got a lot of changes ahead. Everything needs to be centered on ubuntu 16.04 and bootstrap and SSL without exception. Internally we use dockers. but our makefile keeps getting outdated (e.g. docker-ce just switched to overlayfs2 from overlayfs and that isn't a small thing.)

@AdamBark @PaulSD thank you both. we're literally a tiny team trying to balance an elephant on the top of a needle. Your help is SOOO appreciated. My vote is definitely a new branch for 1.10 on 16.04 LTS. 

That said - I always seek the input of our project lead @jennyq as she is quite brilliant and I may be mistaken. Jenny, what are your thoughts?

goetzk:
On 5/3/18 16:24, Ed Schipul wrote:

> As for a major change like the move to 1.10, I believe it definitely
> warrants a new branch. I'm a big proponent of commenting code, moving

In that case, could I suggest the branch be for the 1.11 upgrade and do
it all in one hit?

kk

AdamBark:
My branch is called django-upgrade and I was planning, roughly, one commit per Django version with a final destination of 1.11 as 2.x requires python 3

jennyq:
Hi @AdamBark, thanks for taking the lead on upgrading to django 1.11. I'll make a new branch so that your PRs can be merged to. But before I do that, I'll need to merge another feature branch to master to avoid major conflicts. I'll let you know once it's ready. Thanks again!

PaulSD:
For the record:

[...]

PR #653, opened 2018-03-04
Django upgrade (v1.9)

AdamBark:
To address #613 I've started trying to migrate to Django 1.11 via 1.9 and 1.10. This is step one and shouldn't be merged to master as 1.9 is unsupported but it was suggested in that issue that I submit a PR for each stage.


PaulSD:
Note: There are some merge conflicts between this and #659 ... I can rebase these commits on top of #659 if necessary.

PaulSD:
I've pulled Adam's changes into #659, separated them out into a few different commits for clarity, and fixed a bunch more Django 1.9 issues.

Unless Adam wants to revisit the unused package removal (which should probably be done using a fresh copy from #659 since a lot has changed in that file in that PR), we can probably close this PR.

AdamBark:
It doesn't make much difference to me now, I was just trying to make the upgrade easier but if it's done then the requirements are fine as is.

Issue #439, opened 2015-10-06
fake initials on t7.1 generates future deprecation errors


eschipul:
When running fake initials on an upgrade of 8 it generates numerous future deprecation errors. Adding this issue to track those for milestone tendenci 8 which will be on Django 1.9 (or possibly still 1.8 if 1.9 isn't LTS.) Just capturing the data here.

```
root@tendenci:/var/www/tendenci# python manage.py fake_initials
/usr/local/lib/python2.7/dist-packages/captcha/models.py:36: RemovedInDjango19Warning: Model class captcha.models.CaptchaStore doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class CaptchaStore(models.Model):

/usr/local/lib/python2.7/dist-packages/django/contrib/admin/util.py:7: RemovedInDjango19Warning: The django.contrib.admin.util module has been renamed. Use django.contrib.admin.utils instead.
  "Use django.contrib.admin.utils instead.", RemovedInDjango19Warning)

/usr/local/lib/python2.7/dist-packages/tendenci/apps/events/forms.py:16: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.
  from django.utils.importlib import import_module

Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 351, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 325, in execute
    django.setup()
  File "/usr/local/lib/python2.7/dist-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python2.7/dist-packages/django/apps/registry.py", line 89, in populate
    "duplicates: %s" % app_config.label)
django.core.exceptions.ImproperlyConfigured: Application labels aren't unique, duplicates: testimonials
```

Ran into this error when running the commands here:
https://tendenci.readthedocs.org/en/latest/upgrade/upgrade-to-tendenci7.html#step-5-fake-initials

PaulSD:
Fixed in #659

eschipul:
NA

PR #659, opened 2018-03-11
Tendenci 8

PaulSD:
This PR adds Python 3 support, upgrades to Django 1.11, and fixes lots of unrelated issues.

@aprildh31 and I are also working on upgrading to Bootstrap 4, but that work isn't as far along as the rest of this and still has some major issues, so I've decided to move that to a separate branch for the moment to avoid intermixing too many breaking changes at the same time.

I haven't thoroughly tested everything, so there are probably still some bugs to be found.  However, I have fixed all of the issues I am currently aware of, so this should be good enough for others to start testing.  This code should still be compatible with Python 2, however I have been primarily testing with Python 3, so it is possible I've missed some backward compatibility issues.

For anyone who wants to test, make sure you also use the updated project template: https://github.com/tendenci/tendenci-project-template/tree/tendenci8

Everything currently in the "python3_compatible" branch (and a lot more) is included here, so that branch is now obsolete.

This fixes various issues: #439 #595 #613 #650 #653


jennyq:
Hi @PaulSD, 
Wow, what a PR! I haven't reviewed all commits yet. But I did run your branch on my local. It appears there is a TypeError "method-wrapper object is not an iterator" on content (pages, articles, ...) add/edit. The location of the error is at https://github.com/PaulSD/tendenci/blob/master/tendenci/apps/perms/fields.py#L145. Can you please take a look?

A new branch "tendenci8" has been created. The "entities" branch has been merged to master and will be deleted.  Since this PR also fixes many Python3 compatibility issues and Django 1.9/1.10/1.11 compatibility issues as you pointed out, I think it's better to merge it to the "tendenci8" branch. What do you think?

PaulSD:
Ok, I'll look into that.

There are a few commits here that should definitely go into both Tendenci 7 and Tendenci 8, but most of them can go into Tendenci 8 only if you want.  I'll create a separate PR for the Tendenci 7 commits.

PaulSD:
I've fixed the "method-wrapper object is not an iterator" issue and a bunch of other issues...

PaulSD:
Adding a comment, since I know GitHub doesn't send notifications on PR changes...

I've made a whole lot of updates to this PR, so anyone who is following along might want to take another look at it.

jennyq:
Very much appreciated for getting the ball rolling!
Although I haven't got the chance to finish reviewing all commits yet, let's move forward with the tendenci8 branch!

PR #671, opened 2018-04-02
Tendenci 8

PaulSD:
There are lots of unrelated bug fixes in this PR, but much of this PR is dedicated to extending/enhancing the concept and implementation of themes.

Before this PR, upgrading to Bootstrap 4 in core Tendenci would have been very disruptive to existing users/themes, and upgrading to Bootstrap 4 in a theme would have required effectively forking all of the Tendenci templates and static files into that theme, which would then prevent users from receiving any updates or bug fixes to the base templates and static files after the theme has been installed.

To make themes more powerful, and to facilitate a clean and non-disruptive upgrade to Bootstrap 4, this PR introduces three new theming features:

1. Theme inheritance.  A theme can now "extend" another theme using an "extends" setting in theme.info.  This should allow user themes to be reduced to a much smaller set of customizations on top of an existing theme, enabling users to benefit from updates and bug fixes in the example themes that their user themes are based on.  This has been implemented such that it could also be used to support "mixins" or micro-themes that could be mixed and matched to build a composite theme (by configuring a comma-separated list of themes to extend in the "extends" setting).
2. Implicit static file overrides.  To override static files, you previously needed to override the template that referenced those static files.  This PR enables static files in core Tendenci to be overridden simply by adding them to the theme.  Note that this change requires {{ STATIC_URL }} to be deprecated and {% static %} to be used in its place (as is already recommended by Django).  Templates in existing user themes can continue to use {{ STATIC_URL }} (deprecation warnings will be logged if DEBUG is True), but all templates in core Tendenci and in themes that might be extended must use {% static %} instead.
3. Read-only "built-in" themes.  The standard/example themes are now distributed with the Tendenci pip package.  These themes are read-only in the theme editor, although they can be copied to new user themes via the theme editor for customization.  This gives existing users easy access to the latest versions of the original themes (without their site-specific customizations).  More importantly, this enables users to "extend" the standard themes without modifying them directly, such that they will automatically receive bug fixes and updates to any templates and static files that are not explicitly overridden in their user theme.

With those features in place, Bootstrap 4 support can be implemented in a new "built-in" theme, which can be "extended" by additional standard/example themes, which can be "extended" again or copied and customized by the user to create site-specific themes, all without disrupting any existing user themes.

As a placeholder for Bootstrap 4 work (which is not actually included in this PR), a new built-in "t8-base" theme has been created with a copy of the existing core templates and static files.  For consistency and clarity, the existing core templates and static files have also been moved to a new built-in "t7-base" theme (which existing user themes will automatically "extend"), and the existing standard/example themes have been renamed with a "t7-" prefix to make it clear that they "extend" t7-base and not t8-base.


jennyq:
There are definitely big moves in this PR regarding pulling themes into the core and restructuring core templates and static. I have to admit I'm a little overwhelmed :sweat_smile:.

Some themes like "twenty-thirteen" do not work well with bootstrap and maybe should be removed. We probably should just keep the "tendenci2018" and "creative" themes. I agree keeping the default themes in core will help make them in sync with the core easily. 

Can you please clarify what's the purpose of those static.t7, static.t8, templates.t7, templates.t8, t7-base and t8-base directories?

eschipul:
Paul - this is awesome. Talking through it with Jenny. It's kinda huge as well so wrapping my brain around it. 

I LOVE the fact that it would put everything in the transifex repo and on readthedocs and pip. Just trying to fully grasp the structural changes and migration paths.

And ya, we need a place to chat about this stuff so more people can participate. There is a university in China building out the "django sites" portion for example.

In other words .... one moment please....

jennyq:
Hi @PaulSD, I think I understand the need for those static.t7, static.t8, templates.t7, templates.t8,  t7-base and t8-base directories. Good stuff!

As far as those t7 themes, Ed and I have talked, the only theme we should keep in core is the "tendenci2018" theme. The other themes need to be removed from the core (they either have license issue or not fully bootstrap3 compatible). 

Thank you!

PaulSD:
Yeah, sorry about the giant changes.  I know it's a lot to absorb, but I think this solves a bunch of problems and adds lots of new flexibility all in one fell swoop.

The static.t7, static.t8, templates.t7, and templates.t8 things are simply to make sure developers know where those files have moved, and to ensure that those files are still conveniently accessible during development.  They are not technically required for anything and could be removed.

t7-base and t8-base are the new home for the default templates and static files.  t8-base is currently an identical copy of t7-base, but that will change as t8-base is rewritten for Bootstrap 4.

I'm fine with removing the other themes.  I moved them to the core to consolidate all of the existing themes and clean up tendenci-project-template, but feel free to delete them, or I can do it in my next PR.

PR #674, opened 2018-04-06
Tendenci 8

PaulSD:
A couple more :)


tendenci/apps/memberships/forms.py
```diff
@@ -516,7 +516,7 @@ def assign_fields(form, app_field_objs):
                    if field.field_name != '' and
                    field.field_name in form.fields]
 
-    for name in form.fields:
+    for name in form.fields[:]:
```
jennyq: This would cause "unhashable type: 'slice' error". Googled the issue a bit, it seems the fix is to wrap the list() function around the `form.fields`.

jennyq:
Thank you!

PR #677, opened 2018-04-16
Tendenci 8

PaulSD:
Cherry picking a few commits between themes, and a few new bug fixes.

See also https://github.com/tendenci/tendenci-project-template/pull/17


jennyq:
Thanks!

PR #678, opened 2018-04-16
Fixed some theme issues

replaceafill:
- Updated the image paths in the boxes fixtures to not include `/media/`. It's not necessary now that themes use `static`.
- Removed old `twenty-thirteen` images from the pages fixtures.
- (Not sure I got this right) Updated the sample `content_css` for the `TINYMCE_DEFAULT_CONFIG` setting.
- Fixed the Dashboard links to the theme editor. Using `editor/?file=/...` produces a 404.
- Fixed the favicon name in the base template.


jennyq:
Nice! Thanks!

PR #680, opened 2018-04-26
Fix queryset filter for "disapprove" admin action.

BenSturmfels:
The "approved" and "pending" filters were being and-ed where they should be or-ed, resulting in an always empty queryset, ie. the "disapprove" admin action didn't do anything.

For comparison, see `approve_selected()` on line 92.


jennyq:
Good catch! Thanks!

PR #683, opened 2018-05-09
notifications: Fix spacing in "From:" header causing spamminess.

BenSturmfels:
In our testing, the mail header looks like this:

    From: "Fourth Estate Memberships"<no-reply@fourthestate.email>

Spam Assassin rates this bad spacing as "FROM_MISSP_EH_MATCH 1.441", a significant increase to the overall spam score.

Fixing the spacing reduced the spam score accordingly.


jennyq:
Nice fix!

Issue #531, opened 2016-11-16
Feature Request: Multi Payment Gateways.


FourthEstate:
Please give us option to utilize more than one payment gateway solution. For example Stripe OR Paypal based on the payers preference at checkout.   For a significant enough number of our members they may not be US based, or have a credit card  but they have access to PayPal.

Frankly from a business development and customer experience standpoint giving the member the option for payments that best suits their needs is paramount.

aprildh31:
I agree. Although you can process credit card payments through the right PayPal account type, it would be nice to add a second option for credit cards in addition to PayPal or vice-versa.

FourthEstate:
I'd like to add that I don't think that this feature would be that difficult to accommodate since Tendenci already supports Stripe and PayPal.  The key here is making the choice of payment method select-able as pay using Either Stripe or Paypal on the various payment objects.

goetzk:
This would also be beneficial to us.
@jennyq or @eschipul , would you be open to this being a contributed feature? (of course I can't promise anything at this time...)

FourthEstate:
We would contribute some $ to this effort.

eschipul:
Currency conversion makes this impractical. Workaround would be to create multiple sites. Closing this one - EGS

PR #686, opened 2018-05-30
WIP: memberships: Allow multiple online payment methods in application form.

BenSturmfels:
I'm looking to add multiple payment methods as per #531. This is an initial draft and I wouldn't mind a quick review before I proceed with further testing.

Essentially the approach is that instead of treating online payment as a single thing, the online payment view/URLconf now take a parameter with the `PaymentMethod.machine_name`. To use this, you would set up more than one `PaymentMethod` with `is_online` selected, eg. PayPal and Stripe.

### DIFF
--- a/tendenci/apps/memberships/views.py
+++ b/tendenci/apps/memberships/views.py
@@ -1279,9 +1279,10 @@ def membership_default_add(request, slug='', membership_id=None,
 
             # redirect: payment gateway
             if membership_set.is_paid_online():
+                merchant_account = membership_form2.cleaned_data['payment_method'].machine_name
                 return HttpResponseRedirect(reverse(
                     'payment.pay_online',
-                    args=[invoice.pk, invoice.guid]
+                    args=[merchant_account, invoice.pk, invoice.guid]
                 ))
 
             # redirect: membership edit page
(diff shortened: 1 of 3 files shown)


BenSturmfels:
I was considering removing the "Merchant Account" global setting, but for now this is still needed as a default for `payments.views.pay_online` when no specific merchant account is provided.

jennyq:
Brilliant! Please also test if invoices have any indication of which payment gateway processed the payment.

BenSturmfels:
@jennyq aw, sorry. I'm used to the GitLab feature where putting "WIP" (work in progress) in the title prevents the branch from being merged. Is there a similar feature it GitHub?

The multi payment method feature does essentially work but I'd prefer to test this further in our production system before imposing it on others - especially since it's money-related. I did notice a URL reversing error go by which I'll need to investigate on Monday. I'd suggest reverting for now.

jennyq:
Sorry, I didn't realize WIP means "don't merge". I have installed WIP github app to enable the feature https://github.com/wip/app.

BenSturmfels:
@jennyq my fault, I should have been clearer.

Anyhow, looks like a false alarm on the URL reversing error I mentioned, but I'll leave this a week or so just to make sure no other problems come up.

BenSturmfels:
Hi @jennyq I'd be comfortable seeing this merged in now. Don't know whether that's possible from this closed pull request or not.

jennyq:
Hi @BenSturmfels, Can you please make a new pull request? It's not possible to merge from this one now.

PR #694, opened 2018-06-08
Add scaling information to install instructions

goetzk:
This is information Ed provided to me via email some time ago, tweaked for
presentation here. I'm just trying to make sure this isn't lost, and this looks
like the best fit for the information.


jennyq:
Thanks!

goetzk:
Thanks!

PR #696, opened 2018-06-11
Basic backup and install instructions

goetzk:
Following a question in the tendenci-dev Matrix chat I offered to write up some
documentation to cover backup and restore. This will want testing and
improvements in future but should set people on the right course when doing
their setup.


jennyq:
Nice! Thanks!

goetzk:
Thanks!

PR #705, opened 2018-06-27
Rework install documentation

goetzk:
Below is a complete overhaul of the install documentation, highlights include:
- Removal of (confusing) OS X section
- Combining lots of duplicate instructions
- Removing outdated / unnecessary references
- Updated some section titles
- Fixed capitalisation of Tendenci in numerous places

There is more work to do on this document (like removing any Python 2.7 references), but most of those should be focused at the documentation for Tendenci's Python 3 version.


goetzk:
I will submit a new branch against T11

PR #704, opened 2018-06-25
Document testing requirements and usage

goetzk:
This tries to include the information found in #703 and the documentation
component for #593 .

I don't know if any other work is required as I'm still waiting for my DB setup to complete - unfortunately my dev VM is a touch on the slow side.

PR #707, opened 2018-06-27
Update NoDbTestRunner to support Django 1.8

goetzk:
DjangoTestSuiteRunner is removed in Django 1.8 so NoDbTestRunner
has to be updated to a new testing mechanism.

I'm working on updates to the docs submitted in  #704 which includes
testing the tests.


jennyq:
Thanks!

PR #710, opened 2018-06-29
Documentation sync between T7 and T11

goetzk:
This pulls in all the easy changes I've made between the T11 and T7 branches. The install instructions have been extensively reworked by @PaulSD for T11 so I need to see how my updated docs compare to his and try to sync over only the required changes.


jennyq:
Thanks!

Issue #712, opened 2018-06-29
Is Transifex in use?


goetzk:
Just wondering if the Transifex function in .travis.yml is in use still

jennyq:
Yes, it is still in use. It pushes the message files to Transifex. The message files haven't been refreshed for a while though. But they will be updated for T11.

Issue #708, opened 2018-06-28
Incomplete files migration


goetzk:
Related to #706 (where I had to create files.0005*) I had to update the migration to resolve a schema issue.

```
(testsuite_venv)root@tendenci-combined:/srv/testsuite-tendenci/tendenci_testsite# time coverage run --branch manage.py test --verbosity 2 --keepdb
/usr/local/lib/python2.7/dist-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For detai
ls see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Using existing test database for alias 'default' ('test_tendenci_testsuite')...
Operations to perform:
  Synchronize unmigrated apps: django_admin_bootstrapped, debug_toolbar, tendenci, accounts, formtools, bootstrapform, bootstrap3, gis, api_tasty, tinymce, markdown_deux, djkombu, theme, humanize, haystack, model_report, gunicorn, tags, staticfiles, sitemaps, registry, uploader, mobile, admindocs, messages, social_media, dj_pagination, styled_forms
  Apply all migrations: staff, accountings, redirects, videos, invoices, imports, locations, email_blocks, captcha, navs, meta, explorer_extensions, registration, stories, case_studies, user_groups, contributions, payments, industries, regions, speakers, entities, contacts, event_logs, forums, helpdesk, tastypie, sites, corporate_memberships, pages, forms, discounts, contenttypes, wp_importer, tagging, site_settings, theme_editor, ics, announcements, social_services, events, careers, make_payments, files, studygroups, exports, newsletters, jobs, explorer, sessions, robots, auth, educations, metrics, notifications, boxes, base, campaign_monitor, tendenci_guide, news, wp_exporter, emails, categories, articles, search, committees, handler404, photos, admin, perms, directories, versions, djcelery, social_auth, profiles, recurring_payments, memberships, reports, testimonials, donations, dashboard, resumes, help_files, stripe
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying files.0005_auto_20180626_2307...Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
[...]

Issue #706, opened 2018-06-27
Missing migrations on clean install?


goetzk:
Hi,
I just ran 'manage.py test', and on the second run i had a complaint about missing migrations. Are these missing from the repo or were they missed by django the first time through?

```
root@tendenci-combined:/srv/testsuite-tendenci/tendenci_testsite# python ./manage.py test --keepdb --verbosity 2
/usr/local/lib/python2.7/dist-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Using existing test database for alias 'default' ('test_tendenci_testsuite')...
Operations to perform:
  Synchronize unmigrated apps: django_admin_bootstrapped, debug_toolbar, tendenci, accounts, formtools, bootstrap3, gis, api_tasty, tinymce, djkombu, theme, humanize, haystack, model_report, gunicorn, tags, staticfiles, sitemaps, registry, uploader, mobile, admindocs, messages, social_media, dj_pagination, styled_forms
  Apply all migrations: staff, accountings, redirects, videos, invoices, imports, locations, email_blocks, captcha, navs, meta, explorer_extensions, registration, stories, case_studies, user_groups, contributions, payments, industries, regions, speakers, entities, contacts, event_logs, forums, tastypie, sites, corporate_memberships, pages, forms, discounts, contenttypes, wp_importer, tagging, site_settings, theme_editor, ics, announcements, social_services, events, careers, make_payments, files, studygroups, exports, newsletters, jobs, explorer, sessions, robots, auth, educations, metrics, notifications, boxes, base, campaign_monitor, tendenci_guide, news, wp_exporter, emails, categories, articles, search, committees, handler404, photos, admin, perms, directories, versions, djcelery, social_auth, profiles, recurring_payments, memberships, reports, testimonials, donations, dashboard, resumes, help_files, stripe
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying accountings.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying entities.0001_initial... OK
  Applying announcements.0001_initial... OK
[...]

PR #713, opened 2018-06-30
Use Travis to do a test deploy of Tendenci

goetzk:
To help with the move towards automated testing I wrote 
this travis configuration which does everything except run
the actual Tendenci tests (commented out on line 58).

Because the tests currently fail (on T7 there are 20 failures
and 38 errors) I've left the tests commented out but they're
easy enough to reenable if desired.

Note also that per #712 transifex is still run, this time assuming successful
deployment but that could also be changed to after_script which
would ignore the output of the deployment.

Finally, this is marked WIP because @PaulSD suggested I add
flake8 to the test queue and I haven't done that yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,63 @@
 language: python
 python:
   - "2.7"
-
+ 
+addons:
+  postgresql: "9.3"  
+  # https://github.com/travis-ci/travis-ci/issues/6972
+  apt:
+    packages:
+    - postgresql-9.3-postgis-2.3
+ 
 env:
   global:
+    # Work around boto issue https://github.com/travis-ci/travis-ci/issues/5246#issuecomment-166460882 , https://github.com/tendenci/tendenci/issues/539#issuecomment-345092987
+    - BOTO_CONFIG=/doesnotexist
     # for $TENDENCI_TRANSIFEX_PASSWORD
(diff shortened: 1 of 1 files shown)


goetzk:
No longer marked as WIP - Flake8 can be added in another commit.

jennyq:
This is AWESOME! Setting up a Tendenci site for testing can be resource intensive (that's main reason we didn't do it in the first place). Hope it won't exceed the Travis Cl's resource limit.

.travis.yml
```diff
@@ -1,16 +1,63 @@
...
+ - sudo apt-get -qq install git
+ - sudo apt-get -qq install python-pip
 
 # command to install dependencies
 install:
-  - pip install "Django>=1.8,<1.9"
+ - cd ../..
+ - pwd
+ - pip install -q "Django>=1.8,<1.9"
+ - pip install -e goetzk/tendenci
```
jennyq: Why `pip install -e goetzk/tendenci`? Is "goetzk/tendenci" your github repo?

goetzk: Hi @jennyq ,
It is my GitHub repo, and Travis clones to the name of the repository. I've fixed that in 815a658 where I swap to using a Travis supplied variable.

The current branch is running on Travis at the moment.
https://travis-ci.org/goetzk/tendenci/builds/399389030

goetzk:
The travis test build has passed. Note that this is doing everything short of running the tests. There are two reasons for that:
* Adding them means this is affected by #706 and #708 
* Adding them means builds fail and Transifex doesn't run (which is desirable per #712).

So in a sense, this takes us 90% of the way there and there are just a couple more things to be tweaked before it completely works.

jennyq:
Excellent!

PR #718, opened 2018-07-09
Exclude __pycache__ folder from addons directory

Tusky:
Python creates a dummy directory` __pycache__` which is picked up by the Tendenci addon parser, this is a simple fix, to make sure it's omitted from the list of addons.


jennyq:
Thank you!

PR #735, opened 2018-08-28
memberships: Allow multiple online payment methods in application form.

BenSturmfels:
(This is a new copy of reverted commit c16e8aa relating to #686.)

Instead of assuming a single online payment method configured in settings, this approach passes the relevant `PaymentMethod.machine_name` through to the `pay_online` view. To use this, set up more than one `PaymentMethod` with `is_online` selected, eg. PayPal and Stripe.

If no specific payment method is passed, `pay_online` will use the default specified in the `merchantaccount` setting.

Requirements:

1. Global Merchant Account setting configured, eg. "stripe". Visit: /settings/site/global/#id_merchantaccount

2. Multiple PaymentMethods configured with "is_online" selected. Importantly, use the gateway module name as the "machine name". Can be one of "stripe", "paypal", "authorizenet", "firstdata", "firstdatae4" or "payflowlink". Visit: /admin/payments/paymentmethod/

3. Enable the relevant payment methods on the Membership Application. Visit: /admin/memberships/membershipapp/

### DIFF
--- a/tendenci/apps/memberships/views.py
+++ b/tendenci/apps/memberships/views.py
@@ -1274,9 +1274,10 @@ def membership_default_add(request, slug='', membership_id=None,
 
             # redirect: payment gateway
             if membership_set.is_paid_online():
+                merchant_account = membership_form2.cleaned_data['payment_method'].machine_name
                 return HttpResponseRedirect(reverse(
                     'payment.pay_online',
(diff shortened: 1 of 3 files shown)


jennyq:
Thank you @BenSturmfels !

PR #737, opened 2018-09-06
Pin django-debug-toolbar version

replaceafill:
This PR pins the django-debug-toolbar package to version 1.9.1.

The 1.10 version of the django-debug-toolbar package was [released yesterday](https://django-debug-toolbar.readthedocs.io/en/stable/changes.html) and it's not Django 1.8 compatible which breaks new Tendenci 7 installations.

These are today screenshots from two of our set up scripts showing the problem:

blome59

gtahveq


jennyq:
Thanks!

PR #738, opened 2018-09-14
ascii codec error

martcot:
fix for ascii codec error in apps/theme/template_loaders.py when loading templates with non ascii characters


martcot:
@eschipul  ok it is because it create bug with french accents.

AdamBark:
Just noticed this after posting issue #740. If the default template has non-ascii characters then what you suggest @eschipul doesn't seem helpful.

jennyq:
Since the best solution is to configure your system's default locale to UTF-8, this PR is not necessary.. I'm going to close it. Thanks! 

And I really appreciate @eschipul's security-focused mind which does make us to be more cautious about any possible security ramifications.  Thanks Ed for always staying on top of the security!

PR #741, opened 2018-09-27
Fix stripe request

AdamBark:
In python3 urllib.request.Request will not accept a string for 'data'
so convert to bytes with a UTF-8 encoding


jennyq:
Thanks!

PR #743, opened 2018-10-01
slug validator added to GroupForm

gfbot:
Hello, I was testing your demo tendenci groups, and I found an issues with the **"URL PATH"**, when you try to use **'/'** in the text field and it displays an **"Oops admin error"**. 

screenshot-demo tendenci com-2018 09 25-22-09-02

so, I try to make a fix for that using **_forms.py_** in **user_groups app folder** adding these lines of code

`from tendenci.apps.base.fields import SlugField` **to import tendenci SlugField from tendenci**

**and added this line to class _GroupForm(TendenciBaseForm)_**

`slug = forms.SlugField(label=_("URL Path")) ` 

that works for me but, I'd like to know if you have some advices to fix it properly.

### DIFF
--- a/tendenci/apps/user_groups/forms.py
+++ b/tendenci/apps/user_groups/forms.py
@@ -11,6 +11,7 @@ from tendenci.apps.user_groups.utils import member_choices
 from tendenci.apps.perms.forms import TendenciBaseForm
 from tendenci.apps.entities.models import Entity
 from tendenci.apps.site_settings.utils import get_setting
+from tendenci.apps.base.fields import SlugField
 
 
 SEARCH_CATEGORIES = (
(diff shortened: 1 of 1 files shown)


jennyq:
Hi @gfbot,
I cannot replicate the issue https://demo.tendenci.com/groups/add/. The demo site gets recreated every 30 minutes, it's likely the error was due to the site was being removed and recreated. Have you tried again?

Although the slug field is not specified in the GroupForm, it is specified as `SlugField` for the Group class in the models.py. Thus it does use the validators for `tendenci.apps.base.fields.SlugField`. But I see you changed it to use django's `forms.SlugField` which I think is better. Thanks!

PR #745, opened 2018-10-16
For google auth

TharunThomas8:
After installing the allauth packages the following apps need to be included .

### DIFF
--- a/tendenci/settings.py
+++ b/tendenci/settings.py
@@ -268,6 +268,12 @@ INSTALLED_APPS = [
     # Celery Task System, must stay at the bottom of installed apps
     'kombu.transport.django',
     'djcelery',
+    
+    # For google auth install allauth packages first 
+    # 'allauth',
+    # 'allauth.account',
+    # 'allauth.socialaccount',
+    # 'allauth.socialaccount.providers.google',
 ]
 
 LOGIN_REDIRECT_URL = '/dashboard'


jennyq:
@TharunThomas8, Not sure I understand the purpose of this pull request. Do you just want to make a note by adding those packages that are commented out? They don't need to be in the core. You can include them to your conf/settings.py like this: 

    INSTALLED_APPS += [
     'allauth',
     'allauth.account',
     'allauth.socialaccount',
     'allauth.socialaccount.providers.google',]

https://github.com/tendenci/tendenci-project-template/blob/master/conf/settings.py#L248

TharunThomas8:
I am a django developer and had a project to provide google auth for tendenci.

I had made the google auth for sign up and log in, using allauth ,and was wondering if the allauth libraries could be included in the code base so that I could send a pull request for the remaining set of code.

jennyq:
I think it would be nice. In your implementation, would the login process work for both google auth and regular login? 

@eschipul, what do you think about adding google auth to core? Django-allauth supports many providers, if we have google in place, it would be easy to extend to other providers later if needed, https://django-allauth.readthedocs.io/en/latest/providers.html

eschipul:
@jennyq @TharunThomas8 - I'm all for adding google auth and linkedin auth as well. I'd suggest we make a branch for testing and to see how much additional memory the allauth libraries require in the containers. 

I can also see the possibility of handling it through settings.py, although that could be a bit "brittle" in configuration depending on the load sequence. Maybe, I haven't tested it either. Definitely an overall good thing and I do think we should test it in a branch, via settings or core settings defaults.

Does that work? It seems like an awesome mod by @TharunThomas8 regardless of how exactly it gets implemented.

jennyq:
Good idea @eschipul ! I'll make a new branch for this mod.

jennyq:
Merged it to the branch `google_auth`. Thank you @TharunThomas8! Look forward to testing this mod!

PR #746, opened 2018-10-18
Fix event registration statistics display

AdamBark:
variables seem to go out of scope within 'blocktrans' so tighten the
'blocktrans' to just the pluralisation section.


AdamBark:
After trying a couple of other things I'm not sure it's strictly a scope problem. Using `with event=event` and then using event.spots_available etc. fails but defining each variable in the blocktrans i.e. `with st=spots_taken sa=spots_available` seems to work

Issue #759, opened 2018-11-15
Recurring payments aren't processed in run_nightly


AdamBark:
Stripe doesn't require the MERCHANT_LOGIN and MERCHANT_TXN_KEY variables to be set but if they are the default blank string recurring payments won't be processed by run_nightly
https://github.com/tendenci/tendenci/blob/master/tendenci/apps/base/management/commands/run_nightly_commands.py

I can write a patch but I would like to know how you would like to deal with this, should we just drop the check for these variables? Add to the documentation saying that these should always be set to something if you want to process recurring payments? Add more logic to check for the various sets of variables that should be set for each provider?

jennyq:
A bit of history - 
In Tendenci 4 or early days of T5,  the `MERCHANT_LOGIN` and `MERCHANT_TXN_KEY` were always required to indicate that a merchant account has been set up. Later, some other providers like stripe don't use `login` and `txn_key`. For example, Stripe uses `secret key` and `publishable key`. Apparently, we missed to update the `run_nightly` command afterwards ... Great catch!

PR #760, opened 2018-11-16
Fix recurring_payments test

AdamBark:
The command 'make_recurring_payment_transactions' would only run if
MERCHANT_LOGIN and MERCHANT_TXN_KEY resolved to True (i.e. weren't the
blank string or undefined). This now checks for all the combinations
of variables required by each merchant provider and fixes the bug
where a valid provider has been set up but the payments were not
processed regularly.

### DIFF
--- a/tendenci/apps/base/management/commands/run_nightly_commands.py
+++ b/tendenci/apps/base/management/commands/run_nightly_commands.py
@@ -2,7 +2,6 @@
 
 from django.core.management.base import BaseCommand
 from django.core.management import call_command
-from django.conf import settings
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


AdamBark:
This is a fix for issue #759. I wasn't sure if the right thing to do was to use id_merchantaccount but as it's just a string it seemed easier to misconfigure. Tell me what you think, I can try a different solution if you don't like this one.

AdamBark:
How does this look?

Should this page https://www.tendenci.com/help-files/configure-online-payments-merchant-provider-tendenci/ be updated to remove the unsupported merchants?

jennyq:
It looks good. Thank you @AdamBark ! 

That page is fine. It is for general payments. So, all of those payment gateways are supported. It is the recurring payments that only supports stripe and authorizenet currently.

Issue #725, opened 2018-08-02
Text Not Visible on Membership Page


laealfaro:
As shown in the screenshots below, there is invisible text. I know it is there because I can see it when I highlight the area.

![screenshot from 2018-08-01 21-12-33](https://user-images.githubusercontent.com/42009342/43556824-c0d10db6-95cf-11e8-80dd-9c77af11b9c9.png)

![screenshot from 2018-08-01 21-15-18](https://user-images.githubusercontent.com/42009342/43556885-09fff952-95d0-11e8-90de-e6170710a7eb.png)

PR #769, opened 2019-03-26
Fix memberships details labels

AdamBark:
Class label makes the text white i.e. invisible. Adding a label type —
in this case label-info — adds a coloured background.


AdamBark:
Just for information; it looks like these should be badges in bootstrap 4 as badges and labels are merged to eliminate confusion with the <label> tag. They could be badges now but there's only one style in bootstrap 3.

AdamBark:
This fixes issue #725

jennyq:
Thank you for the fix!

PR #776, opened 2019-04-24
remove schipul.com from defaults

jucajuca:


jennyq:
Thank you @jucajuca !

PR #785, opened 2019-06-03
Update Event Money Output

JustinLewisSMA:
Update to money_collected and money_outstanding properties. Templates
were showing no information on events/search if money_outstanding was
zero. With this change, templates will show more relevant information
on the search page.


jennyq:
Thank you Justin!

JustinLewisSMA:
Thank you!

PR #800, opened 2019-09-02
Tendenci7

dco5:
Fixed but that would not let an officer be deleted from a committee in admin.
Enabled year and month change in datepicker.


jennyq:
Hi @dco5, 

Thank you for your PR. However, Tendenci7 is no longer supported. We only accept PRs in T11 now. 

If you haven't already done so, I highly recommend you to migrate your website to the latest version of Tendenci11. https://tendenci.readthedocs.io/en/latest/upgrade/upgrade-to-tendenci11.html

PR #802, opened 2019-10-02
Lowercase 'SystemD' to 'systemd'

goetzk:
Correcting per the official spelling.
https://www.freedesktop.org/wiki/Software/systemd/

"Yes, it is written systemd, not system D or System D, or even SystemD. And it isn't system d either. Why? Because it's a system daemon, and under Unix/Linux those are in lower case, and get suffixed with a lower case d."


jennyq:
Good to know. Thanks @goetzk !

Issue #811, opened 2019-11-05
Exception Value: 'Email' object has no attribute 'user' sending newsletter


goetzk:
Hi,

I'm told that when attempting to send a newsletter as a non superuser (ie anyone who fails is_superuser) Tendenci generates a 502.

It appears to be caused by this at line 162 of tendenci/apps/emails/models.py
```
    # if this email allows edit by user2_compare                                                                     
    def allow_edit_by(self, user2_compare):
        boo = False
        if user2_compare.profile.is_superuser:                                                                       
            boo = True                                                                                               
        else:   
            if user2_compare == self.user:
                boo = True
            else:
                if user2_compare == self.creator or user2_compare == self.owner:                                     
                    if self.status:
                        boo = True
                else:                                                                                                
                    if user2_compare.has_perm('emails.edit_email', self):                                            
                        if self.status:                                                                              
                            boo = True                                                                               
        return boo
```

Neither the Email or TendenciBaseModel (from which it derives) have a `user` attribute. Neither model has changed substantially in the last 5 years so I am not sure if this is really a long lurking bug or if I've misconfigured the system.

Could you provide some guidance on if this is a bug or configuration issue?

thanks,
Karl.

jennyq:
Resolved https://github.com/tendenci/tendenci/pull/814. Thanks @goetzk again for raising and fixing the issue.

PR #814, opened 2019-11-09
Remove problematic check on self.user (bug #811)

goetzk:
For reasons unknown a check was being performed on self.user, which doesn't exist in the email
class.
Here we remove that check and the associated if statement to allow the code to continue checking. As tested by my customer, this now works for non staff usres. I'm not sure why this broke since a look at the histories for Email and TendenciBaseModel show no changes to add/remove a user attribute for at least 5 years. 

This is resolves Tendenci #811


jennyq:
Thank you @goetzk !

PR #817, opened 2019-12-18
Allow Pages and Stories to appear in apps menu

goetzk:
Currently the Pages and Stories apps are restricted in the Apps navigation menu to
those with superuser access but its not actually required. All users are able
to search - indeed the Pages app displays in the top bar anyway - and should be
able to add pages with appropriate permissions in place.

This change removes the navigational restriction on non superusers.

I haven't had a chance to test this yet so there may be follow up commits.


jennyq:
That makes sense. Thank you @goetzk !

goetzk:
@jennyq Sure, done as #818

PR #818, opened 2019-12-19
Synchronise t7 and t8 _apps_dropdown.html

goetzk:
Based on feedback in #817 I've brought T7 and T8 theme
_top_menu/_apps_dropdown.html files in sync.

In practical terms this means Pages and Stories will show even when the user is
not root and the t8 theme now has a 'Make payment' link in the finances section.


jennyq:
Thanks @goetzk !

PR #820, opened 2019-12-23
Fix visibility of announcements in menu

AdamBark:
Emergency announcements don't appear to rely on the forums module so
remove the test for it.


jennyq:
Nice catch. Thanks @AdamBark !

PR #823, opened 2020-01-01
SECURITY: Prevent unauthorized use of renewal URLs

PaulSD:
Currently, any logged-in user can access any membership by guessing a valid membership application slug and membership ID and using the membership_default.renew URL.  It looks like this behavior was introduced in 26e15c5f.

This commit blocks access to the membership_default.renew URL unless the logged-in user is a superuser or owns the specified membership ID.


jennyq:
Great catch! Thank you @PaulSD !

PR #828, opened 2020-01-26
Remove anonymous user access to payments.

AdamBark:
Remove access for anonymous users to view payment details.


AdamBark:
It seems the anonymous access is there on purpose but I'm not sure why and am concerned it may be a security/privacy issue.

jennyq:
Thanks @AdamBark !

PR #848, opened 2020-04-30
Fixes #847 - .doc (and previously unreported .xls) upload errors

### LINKED ISSUE #847: MIME-Type error when uploading older MSWord .doc document
We're running Tendenci 12.1 on our site.

One of our administrators today tried to upload an older MSWord .doc document about 1.5MB in size.  When we try this we get a MIME-Type error"
```
MIME type 'application/CDFV2' is not valid. Allowed types are: image/gif, image/jpeg, image/png, image/tiff, image/x-ms-bmp, video/x-ms-wmv, video/quicktime, video/mpeg, video/mp4, text/plain, application/msword, text/csv, application/ms-excel, application/ms-powerpoint, application/vnd.ms-powerpoint, application/vnd.openxmlformats-officedocument.presentationml.slideshow, text/vcard, application/pdf, application/zip.
```
When I check this file using mimetype on my computer (Ubuntu 18.04LTS) it returns the expected "application/msword" mimetype, which appears in the list of valid types in the error message.

I had a look at the code that checks the uploaded file (tendenci/tendenci/apps/files/validators.py - FileValidator class) and established that it uses the magic.from_buffer() method to check the mimetype.  I also Googled the error.

From https://github.com/kaleidos/django-validated-file/issues/9 it appears that the magic buffer interface may not be able to determine the correct mimetype for an older Word document unless the whole document is passed to the buffer rather than the 1024 bytes passed by the FileValidator class.  

The above issue from Kaleidos also suggested that reversing the order of testing from mimetype, file size to file size, mimetype would prevent a file size error being incorrectly fed back to the user as a confusing mime type error if a .doc was uploaded

It seems to me the appropriate fix for this would be to reverse the order of mimetype checking and filesize checking and then if the filename extension is ".doc", pass the whole document to the magic buffer interface.

I can have a go at submitting a pull request for this if I get some time over the next day or two.

rob-hills:
Out of the Box Tendenci supports uploading of files with extensions .doc and .xls

However, when these files have the correct mimetypes (application/msword or application/excel respectively), Tendenci incorrectly mis-identifies their mimetype and rejects them with an error that most users will find incomprehensible.

This pull request fixes #847 by:

1. Checking filesize **_before_** mimetype so any file above the site's specified limit will be rejected without having to check its mimetype; and
2. Checking the filename extension and if this is .doc or .xls, feeding the **_whole_** file to the Magic interface so it can correctly evaluate the mimetype.

If the file extension is **_not_** .doc or .xls, only the first kilobyte of the file is checked by the Magic interface as was the case previously.

### DIFF
--- a/tendenci/apps/files/validators.py
+++ b/tendenci/apps/files/validators.py
@@ -53,7 +53,7 @@ class FileValidator(object):
                 raise ValidationError(_("Invalid file name - comma is not allowed."))
             if '..' in value.name:
                 raise ValidationError(_("Invalid file name - two consecutive dots are not allowed."))
-            
+
             # Check the extension
             ext = splitext(value.name)[1].lower()
             if self.allowed_extensions and ext not in self.allowed_extensions:
@@ -64,19 +64,6 @@ class FileValidator(object):
 
                 raise ValidationError(message)
 
-            # Check the content type
-            try:
-                mime_type = magic.from_buffer(value.read(1024), mime=True)
-                if self.allowed_mimetypes and mime_type not in self.allowed_mimetypes:
-                    message = self.mime_message % {
-                        'mimetype': mime_type,
-                        'allowed_mimetypes': ', '.join(self.allowed_mimetypes)
-                    }
-
-                    raise ValidationError(message)
-            except AttributeError:
-                raise ValidationError(_('File type is not valid'))
-
             # Check the file size
             filesize = len(value)
             if self.max_size and filesize > self.max_size:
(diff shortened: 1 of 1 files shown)


jennyq:
Thank you Rob! However, we decided not to merge this PR to the core due to the security vulnerabilities in the old MS Word and Excel that could make .doc and .xls files very dangerous. Thank you for bringing this issue to our attention. We've also removed the .doc and .xls file extensions from the allowed file.

PR #868, opened 2020-06-23
Add tendenci console script for nice install command

iokiwi:
+ Makes the interface for starting a tendenci project simpler
 + More like wagtail
 + Resolve docutils 0.16 dependency conflict with botocore


jennyq:
Nice, a better workflow for the startproject! THANK YOU @iokiwi !

PR #869, opened 2020-06-25
Small fix for python3.8; should still work with earlier

theox26:
Using Python 3.8 encounters an error on deploy, this fixes that error.


jennyq:
Thank you @theox26 !

PR #880, opened 2020-09-02
fix creator_username and owner_username models

evanspaeder:
auth_user table username field is varchar 150 but creator_username and owner_username fields in most apps are varchar 50.  When a user is created with username >150 chars, Tendenci throws a 50* error.  This changes the app fields to 150 characters.


jennyq:
Great catch! Thanks Evan!

eschipul:
Excellent catch! Those are non-relational audit fields intended to discover if a user changes their username, takes an action, and then changes it back. Thus we should be able to search for them in eventlogs accurately using the value used at the time of change.

It stems from an admin who chose a terrible password. A hacker guessed their password and changed the username to be very similar to an entirely different user and then changed it back. Hence, they are not relational fields - display only.

Ideally this data would show for super_users at the bottom of all objects in the admin only box and link to the appropriate search in eventlogs. That functionality hasn't been added back and it's one more thing I need to document as an issue. #todo 

As always, thank you for the merge @jennyq !

PR #881, opened 2020-09-03
Add OpenOffice file formats, remove ppt

### LINKED ISSUE #854: Add support for LibreOffice/OpenOffice Document upload
Tendenci limits file uploads to a number of files, mostly Microsoft application documents, PDF and some image files.

Currently Tendenci doesn't support the upload of documents produced by OpenOffice/LibreOffice applications.  As these are open source office products used in preference to their Microsoft equivalents by many organisations, Tendenci should provide support for these.

Relevant document extensions/mimetypes are detailed in the [Wikipedia Open Document page](https://en.wikipedia.org/wiki/OpenDocument_technical_specification)

evanspaeder:
fix #854
Added OpenOffice / LibreOffice file types - document, spreadsheet, presentation. 

Also removed .ppt.  If we are not allowing .doc and .xls because of potential vulnerabilities / software EOL then it makes sense to also drop support for .ppt.

### DIFF
--- a/tendenci/apps/files/utils.py
+++ b/tendenci/apps/files/utils.py
@@ -540,7 +540,7 @@ def get_max_file_upload_size(file_module=False):
 def get_allowed_upload_file_exts(file_type='other'):
     types = {'image': ('.gif', '.jpeg', '.jpg', '.png', '.tif', '.tiff', '.bmp'),
              'video': ('.wmv', '.mov', '.mpg', '.mp4', '.m4v'),
-             'other': ('.txt', '.docx', '.csv', '.xlsx', '.ppt', '.pptx', '.pps', '.ppsx', '.vcf', '.pdf', '.zip'),
+             'other': ('.txt', '.docx', '.csv', '.xlsx', '.pptx', '.pps', '.ppsx', '.vcf', '.pdf', '.zip', '.odt', '.ods', '.odp'),
              }
     if settings.ALLOW_MP3_UPLOAD:
         types['other'] += ('.mp3',)


jennyq:
Thanks @evanspaeder !

jennyq:
@evanspaeder, Just checked the mime types list in the python mimetypes, those OpenOffice / LibreOffice file extensions are not in there. Which means those file types might still cannot be uploaded. Can you verify please?
https://github.com/python/cpython/blob/3.8/Lib/mimetypes.py#L413

evanspaeder:
@jennyq Yes, you are right. I overlooked that.   

I think we will need to append a list of additional mime types to allowed_mime_types in the get_allowed_mimetypes() function.  I'll submit another PR in a bit.

evanspaeder:
@jennyq  Can you take a look at this commit and let me know if you agree? 
https://github.com/Everblue-Tech/tendenci/commit/07780f7afceb86a54db22a7250e93580ac0df45a

jennyq:
Maybe it's better to check and only add them if they're passed in file_exts, like this?

    od_types_map = {
        '.odt': 'application/vnd.oasis.opendocument.text',
        '.ods': 'application/vnd.oasis.opendocument.spreadsheet',
        '.odp': 'application/vnd.oasis.opendocument.presentation'
        }
    for ext in file_exts:
        if ext in types_map:
            mime_type = types_map[ext]
            if mime_type not in allowed_mimetypes:
                allowed_mimetypes.append(types_map[ext])
        elif ext in od_types_map:
            allowed_mimetypes.append(od_types_map[ext])

evanspaeder:
@jennyq I think that looks good.

jennyq:
Thanks @evanspaeder ! I've picked it up to the commit https://github.com/tendenci/tendenci/commit/5cf42c49553bfefb1efdac4386929c6127d22e70.

PR #894, opened 2020-09-28
fix profile/add form username limits

evanspaeder:
The add member form at /profile/add restricts usernames to 30 characters or less.  This updates the form to match the global username limit of 150 characters.


jennyq:
Thank you @evanspaeder !

PR #897, opened 2020-10-06
fix field lengths in accounts forms

evanspaeder:
Login form username field and password reset form email field did not match global lengths.


jennyq:
Thanks @evanspaeder !

PR #898, opened 2020-10-07
Add additional corporate membership permissions

evanspaeder:
We have a need for non-superuser staff to edit pending corporate membership applications.  This adds an additional permission level for "corporate membership - change pending corporate membership"

It would also be great if we had a permission level to expose the admin only fields to non-superusers.  I haven't dig too far into that yet.


jennyq:
Regarding "expose the admin only fields to non-superusers", it seems that it would need a permission level that allows a non-superuser to act as the module admin. If we have that permission, none of this "change pending" and "approve" permissions would be needed, right?

evanspaeder:
Yes.  I was thinking of making this a separate permission level "view and change admin only fields"

jennyq:
Maybe we should just make it simple -  if they can approve,  they're allowed to edit pending corporate memberships with admin fields included. What do you think?

evanspaeder:
Thats perfectly fine with us

jennyq:
Great! I'll change it.

jennyq:
Resolved. https://github.com/tendenci/tendenci/commit/4f8c4b233e2f0b9088dbc665addee3eeefa9c4dc

Issue #903, opened 2020-10-13
settings_build_init_json fails ... 


bernd-wechner:
Simple enough:

```
$ python manage.py settings_build_init_json
CommandError: Unable to serialize database: a bytes-like object is required, not 'str'
```
Running this after the `deploy` and before `load_tendenci_defaults` in the installation procedure. 

As an aside it would be nice if there were a simple manage.py command like "show_settings".

bernd-wechner:
Fix was merged.

PR #904, opened 2020-10-13
Fix for issue 903

bernd-wechner:
Seems this command didn't keep up with a change in Django's call_command's expectations.


bernd-wechner:
Fixes https://github.com/tendenci/tendenci/issues/903

jennyq:
Nice! Thank you @bernd-wechner !

PR #908, opened 2020-10-15
Add age range option to events

theox26:
This is useful as an information or search point for associations that do specific classes or events for children. If this isn't desired then I'll keep it in my own branch of changes.

I have the example using it in my theme as well if anyone wants that too.


jennyq:
Hi @theox26, 

Thanks for your PR!  If it is to be served as an information or search point, can you simply use event types?

theox26:
I tried tags and they don't allow the formatting I'd need. Types are too limited because they couldn't have the variation we might need and then would be less searchable. 

If this isn't something you want to add to the base project, that's fine. I've already started keeping a branch with my changes since some of them would conflict with your current methods but I need it a specific way.

eschipul:
@theox26 - I recognize the possibilities in this pull request but I don't think this should be merged into core/master/main at this time (cc @jennyq )

My logic is simply that it only helps with search, which could be achieved by event types. I'm unclear how this varies from any other event type or group (committees, study groups, technical interest groups, chapters, etc, are all fundamentally bound to the "group" object which is tied to an entities object. (See [https://www.tendenci.com/help-files/entities/](https://www.tendenci.com/help-files/entities/) for more on that topic.)

Yes, some value. But also danger. I can see this data leading to requiring birth dates and the collection of other personal information which might be against GDPR. Further I can see it being requested to adjust event pricing by the age of the child - which again is what event types already do. Thus I see danger with minimal benefit, even if end users are ultimately 100% responsible to use the software responsibly as explained here [https://gdpr.eu/](https://gdpr.eu/)

Thank you for being a part of the Tendenci community and please don't let my politely declining the pull request discourage you from future contributions. While we are technically a company, we are a very small company behind a much larger community and we would not exist without the help of all of the developers who support our vision and believe in open source and privacy. THANK YOU!

Ed

PR #910, opened 2020-10-16
Added a FAQs view to the help_files app

bernd-wechner:
A much needed feature that the help_files app is missing. 

Addresses:  https://github.com/tendenci/tendenci/issues/911


jennyq:
It's nice to have it. Thanks @bernd-wechner !

PR #925, opened 2020-11-23
fix_import_username_limit

evanspaeder:
User imports are truncating usernames to 30 characters.  This corrects to 150 characters to match the global limit.


jennyq:
Thanks @evanspaeder !

PR #927, opened 2020-11-27
truncate extended for directories/marketplace search results

evanspaeder:
Search results currently leave a lot of empty space in the tiles.  
image

This extends the text to fill the tile.


jennyq:
Thanks @evanspaeder !

Issue #933, opened 2020-12-03
Memberships app: Support timeless dates


bernd-wechner:
I'm working with `joint_dt` and `expire_dt` in Members imports and I see that:

1. they demand time resolution input
2. the display time precision on the Members search page a la (redacted for privacy):

![image](https://user-images.githubusercontent.com/7296506/100956497-ab343e80-356c-11eb-9de1-5bc97101f98c.png)

But a lot of imported data just won't exist with time resolution. I mean if memebership was managed manually and not with on-line registrations for years (as in our case) we recorded dates but not times of day.

A time of 00:00:00 currently renders as "midnight".

Essentially the Memberships module should respect timeless dates. 

A suggestion is:

1. Parse dates flexibly a la: https://github.com/tendenci/tendenci/issues/932
2. Reserve 00:00:00 as a special value to indicate no time (i.e. don't render it as "midnight" just drop time rendering instead.

PR #941, opened 2020-12-04
Support for timeless dates on membership cards

bernd-wechner:
Addressing issue https://github.com/tendenci/tendenci/issues/933


bernd-wechner:
Aaargh got hitchhiking files on the diff again. One of these days I'll work git out ;-).

Issue #937, opened 2020-12-04
Membership Type: Make types more visible


bernd-wechner:
Memberships support typing, and we use this, we have founding member,s life members, supporting members, concession members, family members etc. 

Alas once configured it's actually not easy to see membership types anywhere. This is a braad request to see better visibility of membership types.

I would like at least to see:

1. Type listed when members are listed or wherever Mebership number is shown too.
2. On the Community menu where we Members and Groups also to see Types (Types being as important to us as Groups)

PR #943, opened 2020-12-04
Adding Membership Types to the Community menu

bernd-wechner:
This is a quick hack and proof of concept. I don't much like it, but I think it a mild improvement. Have grouped the Admin entries together at bottom. I would ideally like to see a Membership Types list like the groups list.  

This quick hack addresses the concern of issue 929, that Membership Types need a little more visibility.


bernd-wechner:
Looks good. Getting better at this branching and PRing.

bernd-wechner:
Contributes to https://github.com/tendenci/tendenci/issues/937

jennyq:
Ok, thanks!

Issue #944, opened 2020-12-05
Member Search: Add None to Group filter


bernd-wechner:
There is a problem with member imports in that they do not support any form of group binding (group_id is not supported, but requested in https://github.com/tendenci/tendenci/issues/934).

This means after import it's a manual step to associate the newly imported members with groups. Alas, there is no easy way to find those members (it's looking like a SQL query my end for now). But a minimal need to facilitate it is to find them.

And one way to do that is if on the People/member search page under the Group filter we added support for "None" namely list all members that have no group. 

I may PR such a fix soon (or not). But I certainly need some way to easily move these new imports into groups.

eschipul:
@bernd-wechner Great suggestion. And I see your point. 

Short term, you can do this immediately now in a two step process using /explorer/  and a subsequent import to a catch all group (see below). The help file is here:
[https://www.tendenci.com/help-files/how-use-django-sql-explorer-tendenci/](https://www.tendenci.com/help-files/how-use-django-sql-explorer-tendenci/)

Ideally groupid 1 is associated with entity id 1 and would be immutable and couldn't be deleted but you can't really enforce that at the application level.

As shown in the help file this sql would give you all interactive users. Hence your suggestion.
```
SELECT  u.first_name, u.last_name, u.email, u.username,
          u.is_staff,  u.is_superuser, p.salutation, p.company, 
          p.position_title, p.phone,  p.address, p.address2, 
          p.member_number, p.city, p.state, p.zipcode,  p.country, 
          p.url, p.sex, p.address_type, p.phone2, p.fax,
          p.work_phone,  p.home_phone, p.mobile_phone, p.notes, 
          p.admin_notes 
      FROM auth_user u  
      INNER JOIN profiles_profile p 
         ON u.id=p.user_id 
      WHERE u.is_active=True  
         AND p.status=True 
         AND p.status_detail='active'
```

The data structure I chose in 1999 when I started it is a true EAV model... and while not all of the software adheres to it this is more of a balance of working with the community and not everyone is as adamant about data models as me. However, I would encourage you to also take a look at the intent of entities (money/reporting/scaling) and groups (organization/security/filtering). Typically groups are a subset of a true entities taxonomy.

That help file (get a cup of coffee for this one as it is quite detailed) is located at:

https://www.tendenci.com/help-files/entities/

[...]

PR #946, opened 2020-12-06
Support searching for members NOT in any groups

bernd-wechner:
Improved the quality of the group filter queries too. Discovered the cause of the group_set failure, in that there was ambiguity in it. Creating an explicit related_name, fixed that.


bernd-wechner:
Addresses https://github.com/tendenci/tendenci/issues/944

bernd-wechner:
I noticed uses of user.group_set elsewhere, so did a general refactor of all I could find. it's not extensive. But still, leaves me hoping there is good regression testing in place. group_set worked in code, it failed as a filter field - I presume because of ambiguity given a list of relationships on User included:

```
<ManyToOneRel: user_groups.group>
<ManyToOneRel: user_groups.group>
<ManyToManyRel: user_groups.group>
```
and what appears by inference to be different code paths between the sue of group_set as an attribute in the ORM and as a field name in query filters. Mostly supposition from empirical evidence.

jennyq:
Thank you!

PR #949, opened 2020-12-08
Fixed header on group detail page

bernd-wechner:
It read "Group members in sets of 100" which neither told us how large the group was nor made sense if it held under 100 members. This fixes that.


jennyq:
Thanks!

PR #950, opened 2020-12-08
A small grammatical fix

bernd-wechner:
This menu item read in the singular and opened a form for adding multiple members. Very misleading menu entry the utility of which is discovered by clicking it not by reading it ;-)


jennyq:
Thanks!

PR #951, opened 2020-12-08
Size  member add select boxes more reasonably

bernd-wechner:
Alas these are left to the default select sizing and in CSS what we see happening is Django applying a height (which seems a reasonable height) and then bootstrap overriding that to auto and the size is browser default. 

This is a page specific hack to make the page more usable. What is really needed is a more general think about multiselect boxes, presentation and sizing. The aim here is to improve this page without risking any site wide impacts though a CSS tweak to selectors - though in time that is what I'd recommend.


bernd-wechner:
For the record, all this does is set the size of those multiselect boxes to between 10 and 20 roes based on content (somewhere between if the content lies between).

I do wonder why github wants to throw all that commit noise at us all in PRs. This PR is tiny, one commit, on one branch and it's showing a load of unrelated activity. Odd.

PR #952, opened 2020-12-08
Added some very useful feedback on parsing to import preview

bernd-wechner:
Specifically membership_type (provided as a key, is converted if possible to provide feedbck) and join_dt and expire_dt, are parsed as they will me on import to provide visual preview on successful parse.

Perhaps in future preview columns might be configurable at some level.  These struck me as very useful visuals on my imports to see BEFORE I commit to the import.


tendenci/apps/memberships/views.py
```diff
@@ -568,9 +569,52 @@ def membership_default_import_preview(request, mimport_id,
...
+        # TODO: This could generalize to all ID type imports supported
+        if 'membership_type' in fieldnames:
+            mts = {mt.pk:mt.name for mt in MembershipType.objects.all()}
+            
+            try:
+                mt = int(str(u['membership_type']))
+            except ValueError:
+                mt = 'Value Error'
+                
+            for u in users_list:
```
jennyq, line 616: Looks like this "try... except..." block belongs to the for loop `for u in users_list:`.

bernd-wechner: Indeed. Well spotted. Embarrassing oversight. Working nights ;-)

jennyq:
Great! Thank you!

PR #953, opened 2020-12-09
Memberships Import: A layout improvement suggestion

bernd-wechner:
Having just used this page a lot recently I found myself deeply confused as to what Role, Status and Status detail were. I found no great help on-line either, and so resorted to a code check and it eventuates these are just attached to the group membership record when the members are saved. So I have made a small tweak to the page to make this a more self explanatory behaviour. 

It is not my preference to use in page styling BUT this is a suggestion for now, and triggers a general suggestion to consider consistent styling.and better internal documentation too.

The button for example looked real crowded and it's of style submit-row. Yet that style is not define4d on this page, it is defined only in `admin/css/base.css` whcih isn't included. If I include it the button changes style radically and that is not my intent (to make sweeping styling decisions).

Similarly the iconrow.css file IS included and yet it's not clear why or what styles it provides for what reason on a page like this suggesting it may be a copy./past hitchhiker (not uncommon when starting new page templates on a growing site). I lack great tools for seeing if anything in it is used and where. Something seems used according to Chromes Coverage analysis, but the tools are rigorous in letting me see where by what.

I would ideally like to see Role, Status, Status detail trio in a small invisible table so that the inputs align. I hesitated because tables are a) a pain to style cleanly and b) well out of fashion with CSS grids taking the front seat now BUT a CSS solution is worth doing right not for a little in page patch too.

Even so a table is not totally bad, just that it too would suggest some standard sitewide CSS for properties tables say, that have no borders and consistent styling.


jennyq:
Thanks! 

I think the Status and Status detail fields can be removed because it doesn't make sense to add users to the group as inactive members.  

The icon row is removed from elsewhere, it can be removed as well along with the "css/iconrow.css'.

Issue #954, opened 2020-12-09
Bug: Cannot add committee members


bernd-wechner:
The officer formset has an odd bug I can't pin down quickly. I can reproduce on the demo site:

https://demo.tendenci.com/committees/edit/1/

To do so, create committee, then edit it. You'll land at a URL like the one above.

The try to add committee members. There is a drop down selector for selecting the member and one for the position. 

Click Add Position, as many times as you'd like to create new rows.

You will find it impossible to select a user from rows 2 and on, always the select box in row 1 responds. 

My suspicion is that as the formset is built (new rows added) there is some id confusion (or reuse) so the click event modified the wrong select box. Alternately it might be that the drop down is simply hidden but still there and trapping the click. it seems a bootstrap selector and the element breakdown ad javascript are not trivial and will take more time than I have now to fix, so I just created my committee by adding rows by hand to the database table: `committees_officer`

jennyq:
This issue has been fixed in the tendenci version 12.4.4 and later.
https://github.com/tendenci/tendenci/commit/c3fa2ccf3b2dc776247663a2687789e2e7b8347e

PR #955, opened 2020-12-09
WIP: Bring Group and Officers together on the Committee Edit page

bernd-wechner:
A work in progress. Seeking input.

The Officers can only be selected from the Group that is defined. This binds the controls conceptually. Where it was placed the officers table looks out of place but also it took me some while to work out why I could not find the people I needed on the drop down. It left me confused, and exploring until I noticed the the selection was restricted to the Group! Doh! The UI should ideally be more intuitive, and IMHO selecting a group and then the officers from that group is a sensible workflow, making clear why a Group needs to be selected and on what basis Officers are available.

The User selector is currently broken:

https://github.com/tendenci/tendenci/issues/954

And in any case should be a live one that is responsive to the selected Group.

There is  layout issue on this that I have not reco9lved and hence it remains a WIP (Work In Progress).

The layout issues are 2:

1) The subtle: The Officer table does not align on its left boundary with other fields. It looks slightly out of place. Have not been able to divine why, the amount of CSS at play here is not trivial. 

2) The Gross. There are three boxes at right of the tendenci2020 theme (Get Updates, Site Search and Recent News). The width of the Officers table is such that on a narrow screen (portrait), it overlaps (underlaps) with these. This is a theme design issue and IMHO if at all possible these three boxes should be floating right as they are but wrap around things in the body that need space. in this case they should slip down below the Officers table I guess, Will take some study of the CSS to find a solution there.


bernd-wechner:
I'm not sure exactly what widget is being used for this drop down in Tendenci just yet (need more time to delve). but I should add that I have found django-autocomplete-light to be a wonderful responsive selector.

https://github.com/yourlabs/django-autocomplete-light

Provided with a URL for fetch (in this case the users in a given group) it will when populating its drop down, fetch that list dynamically.

bernd-wechner:
This was a WIP and the PR is messy. Rebased it to main, squashed it, much more tractable but as it's a WIP will close this and keep it local until done.

PR #958, opened 2020-12-15
Support userid = 0

bernd-wechner:
We reserved UserID 0 for the system admin.  It's legal and supported in Postgresql. Works fine. But Tendenci has this one little bug that assumes userids are truthy values. 

There may be other assumptions in code that an ID > 0 but I haven't noticed d any yet, and I think iut fine to start with support but stop short of advocating use of 0 IDs. In short buyer beware, use a 0 userid only if you're game and know how to fix things if need be  - and that is definitely some of us.


bernd-wechner:
Ooops, I just somehow snaggged the chapters branch into these PRs. Please forgive, and if possible just delete PRs #958 to #964 until I work out what on earth I got wrong there and resubmit. Tehe are all about one l or two lines long and easy to triage. BUt got mixed in with the chapters branch. I'll look at it on the morrow.

bernd-wechner:
I found the problem, I merged the chapters branch into my master at some point. BUT I am reading about resets and am confident I can fix these PRs with an appropriate reset. We shall see. Useful reading on git anyhow.

PR #965, opened 2020-12-16
Support userid = 0

bernd-wechner:
We reserved UserID 0 for the system admin. It's legal and supported in Postgresql. Works fine. But Tendenci has this one little bug that assumes userids are truthy values.

There may be other assumptions in code that an ID > 0 but I haven't noticed d any yet, and I think iut fine to start with support but stop short of advocating use of 0 IDs. In short buyer beware, use a 0 userid only if you're game and know how to fix things if need be - and that is definitely some of us.


bernd-wechner:
Now that is better. Alas no real luck with rebasing branches so resubmitted them with a proper parent, and now they look much better (tiny changes). And only the repo owner can do it, but I'd be pleased it the embarassing bungle commits seen in PRs #965 through to  #971 (inclusive) were deleted.

jennyq:
> Now that is better. Alas no real luck with rebasing branches so resubmitted them with a proper parent, and now they look much better (tiny changes). And only the repo owner can do it, but I'd be pleased it the embarassing bungle commits seen in PRs #965 through to #971 (inclusive) were deleted.

I cannot delete them, but can remove them in the merged commits.

jennyq:
Is there any reason why you have to use UserID 0 for the system admin instead of 1?

There is only one line change from `{% if profile.user_id %}` to `{% if profile.user %}`, so it's fine with me.

bernd-wechner:
> Is there any reason why you have to use UserID 0 for the system admin instead of 1?

"have to" would be an exaggeration. "Wish to" is more accurate. It's not a crisis and I'd have left it except i had to redo all the imports so laid out a strategy. During which I noticed we had preexisting member numbers from 1 up. In the first import they imported as userid 2 and up. Not a crisis, but a sore point with some as we use another system here in which the object number (it's a task management system) and the table key are not the same and so in day to day exchanges jobs are referred to by a number but it's impossible to guess the URL from that number. 

Tendenci mostly avoids this by using slugs anyhow so no crisis by any measure. But while there, I just did the quick read on PK 0 in Django and Postgresql and lo and behold others had asked and the answer was "no problem". So I moved the admin superuser (we'll have just one) to ID 0 and imported all the rest from 1 up and member_number and the user_id align now. 

And everything worked fine, but I did notice the admin user conspicuously missing from the user list ;-). So looked into why. Struck me as bad practice to test for existinece with a falsey PK anyhow so I fixed it.

jennyq:
Gotcha. Thanks for the explanation.

PR #967, opened 2020-12-16
A better layout for committee pages

bernd-wechner:
Subjective but I put the committee members after mission before content, and I removed the phone column (who is publishing a committees phone numbers by default today? Easy to add, to a local template but the default should most certainly not assume phone number exposure). And certainly on our COmmittee pages we want to have the mission, then see the members then have the yadayada story about the committee.


jennyq:
I agree that the phone column should be removed. I see you added the title for the list of Officers as "Members", but they are not just members. The real list of members can be pulled from the associated group. As you said it's subjective, the list of officers maybe should be placed further down after the Sponsors or the Notes section.

bernd-wechner:
> I agree that the phone column should be removed. 

Awesome

> I see you added the title for the list of Officers as "Members", but they are not just members. The real list of members can be pulled from the associated group. 

That I suspect is a question of the use case. I can't speak for other users, so leave it for you of course to aggregate your knowledge of how tendenci customers and users operate and how you train them. Our site is now live with the committee page here:

https://hogs.org.au/committees/hogs/

In Australian parlance (and I'm not sure how American or European or other cultures compare) the officers of a committee are those with titles (typically president, vice president, secretary, treasurer, public officer, maybe chair) and the rest are members of the committee. Committees tend to hold under 10 people most certainly under 20, and a page on the committee we want to list the facts up front with the members of the committee listed, then detail to follow.

> As you said it's subjective, the list of officers maybe should be placed further down after the Sponsors or the Notes section.

Up to you. we want it at top as the index, and our Notes introduce them, as you can see form our link (which has a broken image at present alas). In fact we'll go one further as we don't even make member details public, but we do want to make some committee details public and we will modify the links on the table to jump to sections below not to their profiles, as in the public page those links currently just throw up a login prompt which is less than slick. Membership details are presently considered private on our site anyhow, so visible to the committee members and possibly later to other members, but not to the public.

bernd-wechner:
A quick note that I've had a slight hiatus since our site went live. Put two small fixes in recently to help our committee better manage memberships and will PR those in time and finish this one.

bernd-wechner:
Replaced with: https://github.com/tendenci/tendenci/pull/1361 after a tidy up of the PR. Alas github did not update this PR when the source branch was updated. That refers back to here to resolve remaining issues.

PR #968, opened 2020-12-16
Small reordering of code to make more sense

bernd-wechner:
Not a big issue, but form_for_form was being fetched before it was needed and the test following it really is one to be conducted with the tests preceding it. Makes zero functional difference, is just a code style issue.


jennyq:
Ok, thanks!

PR #969, opened 2020-12-16
Removed unnecessary duplication in menu

bernd-wechner:
For some reason two distinct menu items had identical URL. One is clearly redundant. And indeed, Form fields are edited on the Form edit page ... so the duplicate entry can go.


jennyq:
Thanks!

PR #970, opened 2020-12-16
Make email fail mode configurable

bernd-wechner:
To facilitate Email debugging. SIlent failure was hardcoded. Can now be configured in settings.py. Ours were failing and I wanted to see why (i.e. see the exception and stack trace on the console). This makes it possible for anyone.


jennyq:
Cool, it's good for debugging indeed. Thanks!

PR #971, opened 2020-12-16
 Tiny format improvement to Forms menu

bernd-wechner:
Almost insignificant, but a distinct visual improvement to the layout of this menu. It could do with more rework, but then this is true of many of the menus, few of which are as nice say as the Reports menu (nor consistent with it).


jennyq:
This change is not consistent with the other menus. Changed to have two columns, side by side.

bernd-wechner:
Awesome. Consistency is important. Be nice if they all converged on a format as nice as the Reports And Apps menus. In time ...

Issue #972, opened 2020-12-19
Bug: user modified notifications lack username


bernd-wechner:
I just edited some users and I get emails with a subject line:

"User () edited on HoGS"

I've checked the code and that should have the username the brackets there I'm sure, but for now I've not found it easy to trace how and where it gets there. 

I'm willing to be it's from the `user_edited/short.txt` template which contains:

```
{% blocktrans with object.username as objname %}User ({{ objname }}) edited on {{ SITE_GLOBAL_SITEDISPLAYNAME }}{% endblocktrans %}
```
but i haven't found yet precisely where in the code this tenplate is used and `object` is given a value. 

My bet is it's here:

https://github.com/tendenci/tendenci/blob/f99ec9166ea7a076a54bee8d3ac920cb49c24381/tendenci/apps/notifications/models.py#L331

and hence here:

https://github.com/tendenci/tendenci/blob/f99ec9166ea7a076a54bee8d3ac920cb49c24381/tendenci/apps/notifications/models.py#L301

but I haven't yet traced where it gets its `context` from thoroughly. My guess is `object` is missing anyhow.

jennyq:
Great catch! And thank you for fixing it!

PR #973, opened 2020-12-19
Fixed bug 972

bernd-wechner:
See: https://github.com/tendenci/tendenci/issues/972

Maybe not the BEST fix, but certainly the simplest. The lack of a test to see if "object" is provided in the context within send_emails is what allowed this bug to be introduced I suspect. So I'd suggest we also add an assert in send_emails that "object" appears in the context.


jennyq:
Thank you! Since the "object" is not a required variable in the context, putting an assert in the send_emails wouldn't work for other notifications that don't pass in the "object".

PR #974, opened 2020-12-30
Directory creator removed from metadata

robbierobs:
Removed directory creator from the metadata. This information is irrelevant and can be misleading.


jennyq:
Ok, thanks!. It makes sense that directory creator is not always relevant to the metadata.

PR #976, opened 2021-01-22
django_ses.SESBackend conditional added to newletter relay function

robbierobs:
In settings.py, you could had the option to use Amazon SES for the Newsletter backend but there were only checks for the SMTP. If you did not have an SMTP configuration but instead use SES, the send fails and you get the error:
 'Email relay is not configured properly.' 

This commit adds a small check to the `is_newsletter_relay_set function` to see if SES is set (`NEWSLETTER_EMAIL_BACKEND = django_ses.SESBackend` in `settings.py`) and returns true if `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID` are configured in `settings.py.`

### DIFF
--- a/tendenci/apps/newsletters/utils.py
+++ b/tendenci/apps/newsletters/utils.py
@@ -58,9 +58,14 @@ def get_newsletter_connection():
 
 
 def is_newsletter_relay_set():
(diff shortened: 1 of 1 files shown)


jennyq:
Thank you @robbierobs !

PR #978, opened 2021-01-27
More ses fixes

robbierobs:
Added more SES variables to allow django.SES naming conventions as well as using AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID. This also can allow the separation between the two to allow one SES for emails and one for newsletters


jennyq:
Cool. Thanks @robbierobs !

Actually, if these two settings, `AWS_SES_ACCESS_KEY_ID` and `AWS_SES_SECRET_ACCESS_KEY`, are set with django ses backend, both emails and newsletters would use them. This separation is useful "where you would like to use a separate access key to send emails via SES than you would to upload files via S3". https://github.com/django-ses/django-ses

PR #983, opened 2021-02-09
Pending directory added to the directory search

robbierobs:
This allows users with edit permissions or ownership to see pending directories in the directory search. The `/pending` page is too restricted for normal users and this would allow them to edit their directory prior to approval.


evanspaeder:
@jennyq Do you guys have any feedback on this?

eschipul:
@robbierobs @evanspaeder I recognize the business case - 

_"how can a user find their directory listing to edit it further prior to approval?"_ 

Unfortunately this can not be implemented as proposed above. I will offer alternatives after first clarifying why this pull cannot be merged.

### Why it can't be merged? ###
1) The directories search page is, as in all modules, a high volume page which has consistency between between modules. 
2) Any deviation from the basics of "if you are not a superuser then you can't see it" will significantly increase load speed at scale. This will fall down a slippery-slope of a request to view if they are in a particular security group guaranteed.
3) Modules should have consistent naming
4) There is NEVER a time that a user or member can see anything in search unless it is status == True and statusdetail == 'active' (or in some cases statusdetail == 'published'). The point is these are the only two fields that determine visibility without refactoring most of Tendenci.
5) If we change this standard, then every single module including custom modules would need to be changed to use this exact naming convention.

Example: this line introduces ambiguity 
https://github.com/tendenci/tendenci/pull/983/commits/132e69d016b5f4be738554d4ed07c3110ed1b1eb#diff-3343b5a1c18049e4e7b72350f4d54606ffffa3a9e9737cd777dca3851ab7bf2eR82

A quick search for https://github.com/tendenci/tendenci/search?p=5&q=has_perm will show you how frequently and how consistently this standard is applied.

There is too much code above for me to go through each commit and sub-commit line by line.

We don't change core design standards unless you are willing to change them for every single module and every search filter globally. This IS AN OPTION, but one that must be carefully thought out and would be quite time intensive.

### Alternatives ###

Find another way, maybe add a "my submissions" page for a specific user. 
Please compare user.id and not username as username can be edited which is a security hole.
Don't change a global standard without considering the global consequencs.

Lastly - this could be remedied through isolation. KEEP has_perm as is and add a new field as I proposed to Sid. But this method can't be moved into core as it is written.

I will leave this open and welcome any further comments from @jennyq , but as it stands, this introduces inevitable security holes (code is always copied/pasted and some other module will break when another contributor looks at this as an example. 

I will leave this open pending further comments from y'all.But I do know the above code needs to be refactored from ground zero and not change has_perm functionality nor place anything except active content in search. - EGS

jennyq:
Thanks Ed!

Sorry, I wasn't able to respond due to the power outage and lack of internet connection this week since Monday (#texasblackout). 

I agree with Ed on "not change has_perm functionality nor place anything except active content in search". I'll think about the alternatives.

jennyq:
As we discussed with YC during the meeting today, I've added a separate page `/directories/mylist/` as an alternative to display a list of directories (including the pending directories) that are owned by the user https://github.com/tendenci/tendenci/commit/83edcb58fbf244b63414a89b16e41b7b2b0b0d6b. It is linked in the top menu under the Directories (Marketplace).

Issue #990, opened 2021-02-28
Member views: Implement permissions properly


bernd-wechner:
I have a club site and what I expect is a fairly mundane and ordinary expected configuration. That is we have a system admin (IT support) and a committee for the club. The two are distinct. And I was looking at using the finely grained permissions Tendenci supports to open up to the committee the features they need while keeping their menus simple and less cluttered and the system more secure too.

So the committee members are in a group "Committee" and it has permissions. Works a charm.

Imagine my surprise then that having granted them all the membership permissions to find they can't even see members and on users they can't see a lot of data they should be able to.

An immediate problem is visible here:

https://github.com/tendenci/tendenci/blob/master/tendenci/themes/t7-base/templates/profiles/search-result.html

which is full of checks on:

```
request.user.profile.is_superuser
```

and I would like it to read:

```
request.user.profile.is_superuser or request.user.profile.has_permission_for_this
```

In this case I imagine the permission "memberships | Membership | Can view Membership" would be appropriate.

I don't know the syntax off hand for querying permission in a template (hint, please provide some guidance).

But just as badly the Community menu for them shows only Groups and Committees and not Members. The fault for that lies here:

https://github.com/tendenci/tendenci/blob/master/tendenci/themes/t7-base/templates/top_menu/_community_dropdown.html

and this nested set of Iif really should just be one if:

```
{% if MODULE_MEMBERSHIPS_ENABLED %}
{% if USER_IS_SUPERUSER or user.profile.is_member %}
{% if USER_IS_SUPERUSER or MODULE_USERS_ALLOWUSERSEARCH or MODULE_MEMBERSHIPS_MEMBERPROTECTION == 'public' %}
```

though `user.profile.is_member` is a step closer than `request.user.profile.is_superuser`, though still I'd like to know the syntax we could use here to open such menus up to people who have a permission like "memberships | Membership | Can view Membership".

I do understand there is also a naming confusion in this area. That permission presumably by default is referring to memberships (in the abstract, as in what kind of memberships we support, and in our case that's individual, concession and family) as distinct form members and their memberships, but if such reuse of permissions is frowned upon, a quick tip (to save me research and drilling down) on how to add permissions, as any profile that is a member is something a committee needs the ability to manage and should not need to be a superuser to do so (overkill in the extreme)

I'm happy to make these tweaks and PR them if they are a) welcome and b) all the sooner if someone provides some tips on permissions and syntax for template tests on them and conceivably how to add new ones to the fold to use for this.

[...]

PR #996, opened 2021-03-12
Membership View Improvements

bernd-wechner:
Address the issue: https://github.com/tendenci/tendenci/issues/990


jennyq:
Thanks @bernd-wechner ! I'll review it when I get the chance.

jennyq:
Thanks @bernd-wechner !

PR #997, opened 2021-03-12
FAQ ordering fix

bernd-wechner:
A tiny change that fixes the ordering of FAQs. Not ideal, in future I expect it is better to have a sort key as FAQ ordering is something that should be configurable. But this fixes a provisional problem we found adding new FAQs, that they sort high. When we like the order we had and want to add them to end. Which seems more reasonable.


jennyq:
Ok. Thanks!

Issue #1003, opened 2021-03-31
Add Filter ability to Template Tag ListNode


theox26:
https://github.com/tendenci/tendenci/blob/3ee10421305f5afeb1f4dbc0ff02de62c6f20862/tendenci/apps/base/template_tags.py#L35

My actual need is to grab only the featured FAQ from the help_files 'list_help_files' tag. I could make a tag specifically for it or add to the function there but it seems like putting it in this place not only gives me what I need but adds the ability for a lot of other possible uses.

I'll be adding it to my own fork either way, would you like it pushed back here as well?

jennyq:
Hi @theox26, 
If you think it would benefit for everybody else, feel free to submit a pull request. Thanks!

PR #1004, opened 2021-04-01
Tag filters

theox26:
#1003 This allows basic filters in the ListNode class. Only a single operator per group but should cover general use cases.

### DIFF
--- a/tendenci/apps/base/template_tags.py
+++ b/tendenci/apps/base/template_tags.py
@@ -1,6 +1,6 @@
 from builtins import str
 import random
-from operator import or_, and_
+from operator import or_
 from functools import reduce
 
 from django.template import Node, Variable
--- a/tendenci/apps/base/template_tags.py
+++ b/tendenci/apps/base/template_tags.py
@@ -1,6 +1,6 @@
 from builtins import str
 import random
-from operator import or_
+from operator import or_, and_
 from functools import reduce
 
 from django.template import Node, Variable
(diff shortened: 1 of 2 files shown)


jennyq:
Hi @theox26,

Thank you! 

I see it can be useful. However, blindly trusting the end users or devs would do the right things could cause problems down the road, even the security holes in the worst case. It's better to verify rather than relying on the trusts. Therefore, both name and value of the query string variables passed in need to be verified. 

To verify the name of the query string parameters, you can check with the `hasattr`. For example, 

    for k, v in f_qs:
        if hasattr(self.model, k):
            # then do something

To verify the value of the query string parameters, it would need to use the validators for each field. Example,

    field = [field for field in self.model._meta.fields if field.name==k][0]
    try:
        value = field.to_python(value)
        field.run_validators(value)
    except exceptions.ValidationError:
        value = None

    if value is not None:
        # do something

theox26:
Hi @jennyq,

I have updated the code to include the hasattr check, which works great. I see many other methods use that same check within this object. The value validation was unsuccessful in all attempts I made. I don't see a working version of it anywhere and the main example you have above gives a recursion error. Everything else I tried seemed to not be able to overcome calling an attribute from a variable.

I am more than open to discussing it more and trying to figure it out. However, I'm not sure the value validation is as necessary for this use. No other options in the class do it and they are all just as prone to error as this would be. The only use is in the template where a dev should be able to fix their issue without it being a problem for the end users.

Let me know your thoughts.

Thanks!

jennyq:
I've added some verification https://github.com/tendenci/tendenci/pull/1004/commits/2648817e3d19e5d60cc8d77588b506f6fa911a2e. It should be good now.

theox26:
Thank you!

PR #1007, opened 2021-04-04
Translation to pt_BR

farribeiro:
Translation/Localization some sentences to pt_BR


jennyq:
Thank you @farribeiro !

PR #1033, opened 2021-08-13
Remove deprecated CAPTCHA_OUTPUT_FORMAT setting.

BenSturmfels:
Currently when running Tendenci 12.5.8, I get a warning:

```
021-08-13 15:10:17 WARNING py.warnings: /home/ben/.virtualenvs/itpa-django-py38/lib/python3.8/site-packages/captcha/conf/settings.py:34: DeprecationWarning: CAPTCHA_OUTPUT_FORMAT setting is deprecated in favor of widget's template_name.
  warnings.warn(msg, DeprecationWarning)
```

After this change, the CAPTCHA still shows correctly on membership application form, but the warning is no longer present.

(I _think_ my `$PYTHONWARNINGS` are the default, but I have have inadvertently increased the warning level somewhere.)

### DIFF
--- a/tendenci/settings.py
+++ b/tendenci/settings.py
@@ -709,7 +709,6 @@ MERCHANT_ACCOUNT_NAMES = ('stripe', 'authorizenet', 'firstdatae4', 'paypal')
 CAPTCHA_FONT_SIZE = 50
 CAPTCHA_CHALLENGE_FUNCT = 'captcha.helpers.random_char_challenge'
 CAPTCHA_IMAGE_SIZE = (172,80)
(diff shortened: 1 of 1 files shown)


BenSturmfels:
I just discovered that this project does indeed have an increased warning level due to the following changes to `LOGGING`:

```
LOGGING = {
    ...
    'loggers': {
        ...
        'py.warnings': {
            'handlers': ['console'],
        },
    },
}
```

Given that the standard project template doesn't have this logger and the warning doesn't show up with `python -W default` I don't think this is worth pursuing. Closing.

PR #1043, opened 2021-09-20
Added "Export All Members" to Membership export

bernd-wechner:
Membership export was only possible fro subsets of members. A way of exporting all members was missing. A trivial change. 

Two issues with this PR however. 

1) More whitespace change than I like, but that's because my IDE tidies up whitespace. I hope github has a nice diff that can ignore whitespace changes.  if not and it's too hard, drop a line and I'll tidy it up.

2) There is no whitespace around the Export Memberships button. Looks exceedingly ugly. But I've not yet fixed that or worked out the best way as this impacts conceivably many pages. So I flag it. Can provide a screenshot of course as it's hard to imagine it was designed this way and it may be an issue with browsers or platforms. Not likely, but maybe.


bernd-wechner:
Oops 20 commits in this PR. Man Git bothers me.It's a tiny change.

bernd-wechner:
Fixed my IDE to stop formatting on this project and repatched and created a new PR. This one can close.

PR #1044, opened 2021-09-20
Added "Export All Members" to Membership export

bernd-wechner:
Trivial change to add a new option to Memberships export


bernd-wechner:
Never sure why Github lists all those commits ... but the change is trivial.

jennyq:
Thanks @bernd-wechner !

PR #1048, opened 2021-09-21
Added some simple help to two options

bernd-wechner:
I had to consult the online help to work these out. With small help texts conformant with others on same page, I'd have saved myself that effort.


jennyq:
Agreed, it sounds more clear. Thanks!

PR #1049, opened 2021-09-21
Improved expiry widget layout a little

bernd-wechner:
The three day fields were very cryptic and required a help file look up. They are now clear.

The day number fields, rendered very wide with a size 8 attribute. Pulled that in to the required size of 2 (as the day of month has maximally 2 chars) but it was still rendering far too wide, so added an inline style to press the width to 4ch and that renders far more elegantly (2ch is too narrow for some reason).


jennyq:
Yeah, it does look a little better. Thank you @bernd-wechner !

PR #1051, opened 2021-09-28
Small improvements to Member Quick List

bernd-wechner:
As we move our club to rolling memberships, and this list is a central overview tool of our members, I just added an Expiry Date column, and made some minor tweaks to improve this quick list.


tendenci/apps/memberships/views.py
```diff
@@ -2350,7 +2350,32 @@ def report_member_roster(request, template_name='reports/membership_roster.html'
...
+    #
+    # Note the template can render a list without this annotation and
+    # shoudl fall on such a method.  
+    if connection.vendor == 'postgresql':
+        members = members.annotate(user_group_list=StringAgg('user__user_groups__name', ', ', ordering='user__user_groups__id'))
+    
+    if descending:
+        order_by = '-' + order_by
+        
+    members = members.order_by(order_by)
```
jennyq, line 2378: The `order_by` needs to be validated to ensure it contains the intended value before it can be used in the query.

tendenci/themes/t7-base/templates/reports/membership_quick_list.html
```diff
@@ -10,14 +10,28 @@ <h1>{% trans "Current Members Quick List" %}</h1>
     {% trans 'Download CSV File' %}</a></p>
     <table class="table table-tendenci-reports">
         <tr>
-            <th>{% trans "Last Name" %}</th>
-            <th>{% trans "First Name" %}</th>
-            <th>{% trans "Company" %}</th>
+            <th><a href="{% order_by request order_by 'user__last_name' %}">{% trans "Last Name" %}</a></th>
```
jennyq, line 13: Where is this `order_by` template tag coming from?

tendenci/apps/memberships/views.py
```diff
@@ -2350,7 +2350,32 @@ def report_member_roster(request, template_name='reports/membership_roster.html'
...
+    order_by = request.GET.get('order_by', 'user__last_name')
+    
+    descending = order_by.startswith('-')
+    if descending:
+        order_by =  order_by[1:]
+    
+    members = MembershipDefault.objects.filter(status=1, status_detail='active')
+    
+    if order_by == 'member_number_int':
+        members = members.annotate(member_number_int=Cast('member_number', PositiveIntegerField()))
```
jennyq, line 2362: The `member_number` is a char field. Trying to convert it to an integer wouldn't always work.

tendenci/apps/memberships/views.py
```diff
@@ -2350,7 +2350,32 @@ def report_member_roster(request, template_name='reports/membership_roster.html'
...
+
+    # If we want to be able order by an aggregated list of groups
+    # then we need to annotate the with that. Django provides no 
+    # generic means for creating such aggregates yet but the postgresql
+    # package provides one. So we can order by the group list only on
+    # postgresql systems for now. 
+    #
+    # Note the template can render a list without this annotation and
+    # shoudl fall on such a method.  
+    if connection.vendor == 'postgresql':
```
jennyq, line 2372: Order by user groups list? That could cause performance issues and it doesn't seem to be useful.

jennyq:
Hi @bernd-wechner, 

Thanks for the PR. I've added some comments. Please review. Thanks!

tendenci/apps/memberships/views.py
bernd-wechner: Oops, forgot to stage and commit the tag. Fixed.

tendenci/themes/t7-base/templates/reports/membership_quick_list.html
bernd-wechner: Oops, forgot to stage and commit the tag. Fixed.

tendenci/apps/memberships/views.py
bernd-wechner: I find it very useful ;-).

tendenci/apps/memberships/views.py
bernd-wechner: Good point. Made it optional. Essentially the default template will sort by `member_number` and a local template (override) can implement `member_number_int` for associations that use exclusively int member numbers. I am guessing it was a string to support alphanumeric numbers (akin to car registration plates), but it is a fair assumption that many, like us, have no truck for that and only want to use numbers ... they (including us) can tweak the template locally by adding `_int` to the template tag in the right spot (we will).

tendenci/apps/memberships/views.py
bernd-wechner: Have you any tips on how? Am struggling to find a validator for field lookups at present. Basically fields can be on membership or membership.user or even deeper in theory, but there's a syntax for field lookups such that `user__lastname` is a reference to a field on `membership.user`. I'd like to think a general purpous validator for such field lookups already exists, but I can't find one yet.

PR #1053, opened 2021-09-28
Forms help text improvement

bernd-wechner:
Had to consult help to understand these fields again. That step saved with a small help text fix.


jennyq:
Thanks @bernd-wechner !

PR #1056, opened 2021-09-28
Use logged in users email as fallback for form submissions

bernd-wechner:
I just created an expense claim form for our club. It's private and only available to people with accounts and you need to be logged in to use it. In spite of the forms configuration, people are not getting emails. Looked into why. And this is why.

On such a form I don't want to add an extra field, visible or hidden, that contains the users email address (both unnecessary and exposes email addresses to clients).

To wit, I think it sensible to check the logged in users email address and use this as a fallback.

My one assumption here is that self.user always has a value (even if it's an anonymous user) and I make no assumption that the email attribute exists or has a sensible value.


tendenci/apps/forms_builder/forms/forms.py
```diff
@@ -258,6 +258,11 @@ def email_to(self):
             field_class = field.field_type.split("/")[0]
             if field_class == "EmailVerificationField":
                 return self.cleaned_data["field_%s" % field.id]
+            
+        user_email = getattr(self.user, "email", None).strip()
```
jennyq, line 262: In case of anonymous user, `getattr(self.user, "email", None)` would return `None`.  It would then throw an AttributeError error because `None` is not a string.

bernd-wechner: So: `user_email = getattr(self.user, "email", "").strip()` is preferred?

jennyq: Yes, that would be good.

bernd-wechner: WIll fix and push to branch.

PR #1057, opened 2021-09-29
Added support of a "groups" field on Membership imports

bernd-wechner:
Membership import remains important to us. And we use groups extensively. To wit, we wish to support a simple way of importing the groups a member is in. 

I took inspiration from the existing clean_membership_type() methods, to spin of a clean_groups() method which similarly supports integer IDs or names but distinctly from membership_type a CSV list of values (as a member can be in multiple groups. Cleaning produces a list of valid group IDs.

Then when executing the import we honour it is cleaned field if it is provided, else fall back on the legacy behaviour of attaching the default group based on membership type.


tendenci/apps/memberships/views.py
```diff
@@ -552,9 +552,9 @@ def membership_default_import_preview(request, mimport_id,
                 
         # DateTime fields are sensitive to parse failures
         # They are not parsed in the preview yet, in fact all 
-        # data travens as strings to be cleaned and parsd just 
-        # before being saved to the respecive models. Datetimes 
-        # and dates are parsed with dateutil.parser, so we fo
+        # data travels as strings to be cleaned and parsed just 
+        # before being saved to the respective models. Datetimes 
+        # and dates are parsed with dateutil.parser, so we do
```
bernd-wechner, line 557: Simple typos in a comment fixed.

tendenci/apps/memberships/utils.py
```diff
@@ -1211,15 +1217,12 @@ def clean_membership_type(self, memb_data, **kwargs):
                 else:
                     memb_data['membership_type'] = value
             else:
-                if not MembershipType.objects.filter(
-                                            name=value).exists():
+                candidates = MembershipType.objects.filter(name=value)
+                if not candidates.exists():
```
bernd-wechner, line 1221: Simple DRY improvement. 
We keep a copy of `MembershipType.objects.filter(name=value)` in `candidates` for reuse in the `else` block rather than recoding the query.

bernd-wechner:
Over a coffee I marked up all the changes to facilitate an efficient review and merge (practicing some github interaction really).

bernd-wechner:
Closed and moved to https://github.com/tendenci/tendenci/pull/1359

PR #1060, opened 2021-09-30
Form entry lists have a configurable summary

bernd-wechner:
Required addition of a model field and hence a migration. 
Form designer can configure the summary card format with which form entries are displayed.
The format of this new text field would need documenting. Can do that too on request of course.


bernd-wechner:
I'm loving this already. All out form entries are now so so more legible when committee members check them.

bernd-wechner:
I hadn't noticed the `position` field, it's not in the UI, but if it's the database, good point, I will check when next able. The legacy break I had already noticed and will fix (that's easy to fix). It would also need documenting but that help page needs fixing anyhow and I'm happy to do that. It doesn't describe the available fields well enough leaving us to experiment with them alas.

As to how to use use I can illustrate my expense form:

image

which produces summaries like:

image

The main and huge improvement in workflow for us being I see the sum claimed, what for and how the claimant requests us to compensate them AND not in this sample, if we've paid them the first word on second line is Paid.

The syntax for the Summary Position fields is as follows:

1. It's a CSV string
2. format is either:
    1. row, col
    2. col
    4. row, col, fmt
    5. col, fmt
3.  where:
    1.  row is a number specifying the row starting from 1. Missing rows are collapsed 
    2. col is the column with columns separated with " - " presently
    3. if no row is specified 1 is assumed
    4. if 1,1 is empty then the submission date time is formatted there (as per legacy)
    5. if row 1 is empty it (will, this is todo) produce the legacy single line
    3. fmt is an optional format string, that currently accepts:
        1. $, which formats it as a currency (todo: localisation, just uses "$" sign presently.
        2. w<n>, where n is an int, which prints the first n words (a legacy support format being w2)
        3. b, where the field is boolean, it will not write True/False but FieldName/NOT Fieldname (intended for my use as Paid, NOT Paid.  
        4. Of course this is extensible.

Of course if there's an existing field and it's not used in the UI that would save a migration too. As noted will check when next able. Just a quick reply while I'm online for now (but away from that workstation).

bernd-wechner:
OK:

1. The `position` field is for positioning the field on the submission form and is a sort key, simple ordering value.. The new `summary_position` specifies the position of the field on the summary card on a list of entries and has a 3 value CSV syntax.
    -  format described earlier, but I added B now for Basenaming files.
2. I reimplemented legacy behaviour when no summary positions are specified. That is, existing users will see no change in behaviour on their forms.
     -  This is of course now all implemented in the `summary` property of `FormEntry` and not in the template as it was. 
3. I fixed a bug (saving fields on edits was failing - I was sure I'd tested that thoroughly and had it working but somehow, a small typo seems to have prevented it from working in my develop branch.

tendenci/apps/forms_builder/forms/models.py
```diff
@@ -234,6 +236,8 @@ class Field(OrderingBaseModel):
         help_text=_("Comma separated options where applicable"))
     default = models.CharField(_("Default"), max_length=1000, blank=True,
         help_text=_("Default value of the field"))
+    summary_position = models.CharField(_("Summary Position"), max_length=6, blank=True,
+        help_text=_("Position in the onle line form entry summary. Row, Position or just Position."))
```
jennyq, line 240: 1) "onle line" or "one line"?

2) Can you please add some examples here so that users don't have to guess what can be entered.

3) The value of the `summary_position` field (entered by users) needs to be validated. This can be done by either adding a [validator](https://docs.djangoproject.com/en/3.2/ref/forms/validation/#using-validators) for the field `summary_position`, or adding a method `clean_summary_position` for the class `FormForField` in [forms.py](https://github.com/tendenci/tendenci/blob/master/tendenci/apps/forms_builder/forms/forms.py#L467) to throw [ValidationError](https://docs.djangoproject.com/en/3.2/ref/forms/validation/) if there is a problem with the field.

bernd-wechner: Thanks! All good ideas. Will do.

ndavis2: Hi Bernd,

Checking in to see what the outcome was of this request. I look forward to hearing from you!

Kind regards,
Nicole, Team Tendenci

bernd-wechner: > Hi Bernd,
> 
> Checking in to see what the outcome was of this request. I look forward to hearing from you!
> 
> Kind regards, Nicole, Team Tendenci

Dropped the ball on this. Good reminder. Will leave the page open to take a look when I can squeeze the time in again. The validator is the thing that will take a moment to write, but is well worth doing indeed.

bernd-wechner: Just fixed these three, but got the validator wrong. I just realised I have to write one after all ;-). And up the help a little.

bernd-wechner: And now I fixed the validator. It provides rather comprehensive feedback, too.

jennyq:
Looks great now. A big thank you @bernd-wechner for bringing in this feature!

ndavis2:
Can't wait to see this in action - a huge thanks @bernd-wechner !!

-Nicole, Team Tendenci

ndavis2:
Thanks @jennyq for reviewing and merging this feature!

lashirley21:
That's a great expense summary. Thank you @bernd-wechner and all.

Issue #1058, opened 2021-09-29
Forms: Add option to edit form entry


bernd-wechner:
I've created an Expense claim form, for committee members. It's working well. They simply upload a receipt and nominate a payment means and as treasurer I approve and pay the expense. I added two hidden unrequired fields **Paid** and **Paid On**, but I need to be able to edit the entry not just view it, to check the Paid checkbox and select a date. This is a simple workflow requirement on some form types, that the submission can be updated after submission with some status (like done).

PR #1062, opened 2021-09-30
Initial cut at form entry editing capabilities. 

bernd-wechner:
Looks robust to me, plays conservative where possible and maintains all backward compatibility, while permitting form entries to be edited. Specifically addressing the issue: https://github.com/tendenci/tendenci/issues/1058


jennyq:
Yes, django has those 4 perms - add, change, delete, and view - created by default for each model. 

Okay, I see your point. With the `editable` boolean field added to the form fields and permission check on the entry edit, we should be good. 

    if entry:
        if not has_perm(request.user,'forms.change_formentry', entry):
            raise Http403

bernd-wechner:
Let me know what I can do to see this merged. I've been away a while.

jennyq:
Hi @bernd-wechner,

The `has_perm` function takes `user` and `perm`, or `user`, `perm` and `object` as arguments. https://github.com/tendenci/tendenci/blob/master/tendenci/apps/perms/utils.py#L155 (the actual implementation is [here](https://github.com/tendenci/tendenci/blob/master/tendenci/apps/perms/backend.py#L110)). If `object` is present, it should match with the type of the `perm`.  For example, we cannot mix `articles.view_article` perm with `page` object. Likewise, we don't want to mix `forms.view_formentry` perm with `entry.form` (that's a form object).

So, this is correct:

    has_perm(request.user,'forms.change_formentry', entry)

This is incorrect:

    has_perm(request.user,'forms.view_formentry',entry.form)

Does that make sense?

I suggest to apply `has_perm(request.user,'forms.view_formentry',entry)` only to the `entry_detail` view, and restore others back to their original format.

bernd-wechner:
Thanks Jenny,  I'll take a closer look as time permits. Your feedback is always appreciated.

bernd-wechner:
Jenny,

Made time for this finally. On review:

1. I'm keen to keep permission relevant to the context, and have endeavoured to do that now
2. I have kept the object consistent with the permission as per your useful advice there.
3. I have remove reference to an object where no object is relevant.

I hope that passes muster, and as ever am keen to progress this. If it's OK, I'll look at rebasing it to the latest commit and then resolving any conflics.

jennyq:
Thanks Bernd!

Two more things:
1) On line 377, the form assignment is not feasible or efficient.

       form = get_object_or_404(published, entries=entry)

   With entry, form can be assigned simply with entry.form.

       form = entry.form

2) On line 386, please change `has_view_perm` to `has_perm`. The `has_view_perm` is used in details view and the model of the object needs to be a subclass of TendenciBaseModel. 

Other than that, it looks good.

bernd-wechner:
So, set to rebase it from latest commit then?

jennyq:
Yes, please move forward to resolve the conflicts.

jennyq:
Thanks Bernd for adding the edit feature for form entries!  Great work!

bernd-wechner:
Not at all, thank you very much for your kind support!

Also, scratching my head, how that last little change you made was needed. My local production branch (our live site) has it exactly as you changed it to, and somehow, from the first commit on the branch of this PR that one line differed. And I never caught it! To wit, incomplete testing on my part of the branch! (fully tested prior to branching a PR, but somehow this line got broken. Left wondering if it's an artefact of my efforts to juggle three versions of forms.py on three branches.

The other one that's been overlapping with this one a lot and causing me headaches because of it is:

https://github.com/tendenci/tendenci/pull/1114

Which I will shortly rebase to the master head, resolve any conflicts and test properly (I have in fact been developing that on the master branch anyhow, not on my local production branch, and so I don't need to cherry-pick commits).

jennyq:
Yeah, that line could be overlooked while resolving merge conflicts. Simultaneously working on different feature branches in the same module can be tough.

PR #1063, opened 2021-10-01
Redirect after From Entry Deletion to List of Entries

bernd-wechner:
It was redirecting to the list of forms. Nor a comfortable workflow. I'm deleting entries and so want to return to the list of entries.


jennyq:
Thanks @bernd-wechner !

PR #1064, opened 2021-10-01
Grant staff access to reports

bernd-wechner:
I'm stuck on this. It needs more work to be sure, and more nuanced implementation. I see no permissions related to reports. Immediate need I have is that the committee on this association have access to membership reports.

Alas the granularity is not great as when I try it there are Report items I have no permission for and am informed only when I try to use them. That's actually fine, but the menu should not show them. The granularity is currently lacking.

Also the roles are not granular enough. Essentially we need report permissions to associate with groups. As I have a Committee group. Staff is broader, could include folk we really don't need to expose membership data to. Staff are volunteers doing work for us, or if we were so lucky we hired ;-)

But if implemented using the perms app then if could be broken down to user, membership, financial and site report permissions with their audiences (user and site, system admins, membership, users and financial committee.


bernd-wechner:
Closing and moved to https://github.com/tendenci/tendenci/pull/1358 because of Githubs's refusal to propagate branch updates to the PR.

PR #1067, opened 2021-10-10
Add form memory

bernd-wechner:
Not least because of COVID tracing requirements we use an attendance registry form.  It asks attendees for their name, email and phone number and they can scan a QR code at events and up pops the form in their browser. 

We run weekly events and every week this form is rendered empty. 

On desktop browsers because the `name` attribute is set you can click in the field and see last weeks entry. Alas on mobiles not the case on tested browsers. 

To wit, have added another form field attribute called `remember`, which if set stores the last saved value int he Django session and when the form is loaded initialises the widget with the session stored value. 

Passes internal tests and is live. 

The intent is that as a regular you can scan the QR code and see the form again but from week 1 only need to click Save to register again. The time of course is stored with he save and tells us what event they were at.


bernd-wechner:
Made a schmozzle of this branch so deleted it, and wiil start again.  It mixed in code from other branches not merged into main, so needs a rebase first.

PR #1068, opened 2021-10-10
DRY the field key in FormForForm

bernd-wechner:
A simply DRY improvement is all. Pedantic but hey.


bernd-wechner:
Also adds support for global keys (which I have plans for ;-).

tendenci/apps/forms_builder/forms/forms.py
```diff
@@ -188,6 +188,12 @@ def add_pricing_fields(form, formforform):
 
         self.add_form_control_class()
 
+    def field_key(self, field, is_global=False):
```
jennyq, line 191: Where is the use case for the `is_global` variable?

PR #1069, opened 2021-10-10
Add form memory

bernd-wechner:
A clean PR this time. A small change that provides for forms with memory. 

We need this for user convenience, some of our forms especially an attendance registry is used repeatedly with simple new submissions. Essentially forms lack memory to date. They dos et the name attribute on fields and on desktop browsers that makes some history available. Alas on mobile browsers checked, not. Essentially we want one of our visitors to open our attendance register and if they submitted one last week, they now just have to click submit, the form is filled for them as it was last week.

This saves any fields flagged with `remember` in the session. Django keeps anonymous and logged in sessions and this is configurable with regards to expiry times on these sessions. This one is cleanly mergeable, but interacts with the form editing PR, I had to separate clean mast branches because of these interactions.


bernd-wechner:
This is working admirably, and I've added a quick dirty view to watch how memories are evolving.

jennyq:
I'll look into the merge conflicts when I get the chance.

bernd-wechner:
Kind of keen to see this move ...

tendenci/apps/forms_builder/forms/urls.py
```diff
@@ -17,6 +17,11 @@
     re_path(r'^%s/entries/export/(?P<task_id>[-\w]+)/check/$' % urlpath, views.entries_export_check, name="form_entries_export_check"),
     re_path(r'^%s/entries/export/(?P<task_id>[-\w]+)/download/$' % urlpath, views.entries_export_download, name="form_entries_export_download"),
 
+    re_path(r"^%s/entry/edit/(?P<id>\d+)$" % urlpath, views.form_detail, name="form_entry_edit"),
```
jennyq, line 20: I don't see `views.form_detail` handles entry edit. Besides, it accepts "slug",  not "id", as the parameter. Did you forget to implement it?

jennyq:
Hi @bernd-wechner,

The conflicts have been resolved. 
Did you forget to implement the entry edit? If it is not intended, can you please remove the url pattern for "form_entry_edit"? Thanks!

bernd-wechner:
And fixed!

jennyq:
Thanks @bernd-wechner !

tendenci/apps/forms_builder/forms/views.py
```diff
@@ -243,6 +245,37 @@ def entries(request, id, template_name="forms/entries.html"):
...
+def memories(request, id, template_name="forms/memories.html"):
+    form = get_object_or_404(Form, pk=id)
+
+    if not has_perm(request.user,'forms.view_formentry',form):
+        raise Http403
+    
+    key_pattern = re.compile(f"{form.slug}.field_(?P<field_id>\d+)")
+
+    # Not sure how well this scales and if perhaps a server side query (filter) is better 
+    s_mem = [s for s in Session.objects.all()
```
jennyq, line 259: Hmm... your comment about the scalability here is right - It will NOT scale. A possible better way is to have another table to store all the session keys whose session data contain forms data, then it can filter the session keys when querying the Session table. For now, I'm going to remove the `memories` url. 

Another possible issue regarding form memories: If the form data exceed the [common limit of 4096 bytes](https://tools.ietf.org/html/rfc2965.html#section-5.3) per cookie, browser can not store all the data and the session data will be invalidated. The result would be users wouldn't be able to log in without deleting the cookie from their browser. The form memories can only work for small forms.

bernd-wechner: If I understand right though, this adds nothing to cookie memory requirements. The form data is not stored in the cookie in an ordinary Django configuration. I will have to check if that's a universal Django truth or a configuration dependent one.

Much rather Django stores a session key in a cookie and has a Session table in which the session data is stored.  For all the reasons you point out. 

On one site run I actually store rather large quantities of cache information in the Session store. It's a convenient way to store expensive results (the results of queries).  Django offers dict based caching, and memcached offers another option but the beauty of the Session store is the cache is persistent across memory refreshes (reboots, server restarts etc). So the choice of caching store depends on the need.

My concern about scaling there is that  this creates a list of Session objects, which can be large as there is conceivably one Session object per session, where a session is any logged in or not visitor with a session key stored as a cookie. The query approach I mooted as likely better avoids making a list of all sessions but creates a query (SQL) that  returns only sessions that contain form memories for this form. So the scaling issue is not form size much rather user community size (people using a given site).

jennyq: I was overthinking. You're right - session data don't get passed to browsers, session keys do.

If we assume the old sessions get removed regularly (the `clearsessions` is included in the `run_nightly_commands`), it is true that the size of the session's table is determined by the number of site visitors. So, if a site has ten thousands of members, it would have at least ten thousands of sessions. The operation to decode each session and find a match would still be very expensive. The query approach sounds good, but how do we generate the query without decoding every session? Why don't we use additional table to store form memories? The table can have these 3 columns: session_id, form_id, data (a dictionary).

bernd-wechner: Good points, and I think it worth moving to a new model. I'll try and fix that and PR it this week as I'm stuck at home in quarantine this week, thanks to a child testing COVID positive, so a lot of dun dancing around one another not allowed to leave the house ... Have not got much done since it started but imagine I may find a chance to fix this before it's done. One week we have mandated quarantine at home here when a family member tests positive. Fortunately he's only 4 and not very ill at all, but still them's the rules here.

jennyq: Thanks @bernd-wechner ! Sorry to hear that you have a child testing covid positive. Take care and stay safe!

bernd-wechner: Small change of plan as my workstation went down in sympathy. Dead. Have ordered a new motherboard, and won't have a work station till that arrives (about 2 weeks nominally). Spent this morning shopping and there is little I like less.

jennyq: No worries.

bernd-wechner: I'll probably get this fixed by end of May by looks of it now as I wait for parts on my workstation. In the interim a lot of the non IT projects around here will get some of my attention.

bernd-wechner: Am moving this to a new model now. A little tricker than hoped as I have to add a migration tool (given it's live) from the old session store to the new model. No big drama, just another step ;-)

bernd-wechner: And move to model is done in: https://github.com/tendenci/tendenci/pull/1114

Issue #1083, opened 2022-01-15
14.0.1 -> 14.0.2 upgrade pytz version conflict due to Celery upgrade


rob-hills:
Hi,

I just tried to step through upgrading our test server from Tendenci 14.0.1 to 14.0.2 but encountered a package dependency conflict.  This was the error message:

```
ERROR: Cannot install tendenci and tendenci==14.0.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    tendenci 14.0.2 depends on pytz==2021.1
    celery 5.2.3 depends on pytz>=2021.3
```

I'll submit a pull request with the fix that worked for me.

PR #1084, opened 2022-01-15
Update requirements.txt

rob-hills:
Upgrade Pytz version to fix dependency conflict resulting from Celery upgrade to version 5.2.3 (#1083)


jennyq:
Thanks @rob-hills !

PR #1091, opened 2022-02-08
Highlight system file changes in Upgrade

rob-hills:
System file changes required after VirtualEnv upgrade.  Improved visibility of the documentation about this.


jennyq:
Thanks @rob-hills !

PR #1094, opened 2022-02-28
fix: pinning django-countries due to 7.3.0 update

robbierobs:
A recent update to `django-countries` (version: 7.3.0) breaks the Tendenci installation. Pinning `django-countries` to the previous version resolves this issue (version: 7.2.1).


jennyq:
Thanks @robbierobs ! Closing it as the issue has been resolved already.

PR #1101, opened 2022-04-01
Add Square as a Payment Processor option

theox26:
This has all back-end code updated to allow Square as a payment processor. It also allows Square gift cards to be validated and used. An ajax method to check gift card balance is included. The method allows for a gift card partial payment then the rest paid using a credit card.

If another processor was used in recurring payment previously, it allows the old processor to be used until the account is updated.

My theme is very different from the base theme and would be the only known issue. I can break that out in a separate set of commits but this shows the usage in full.


jennyq:
@theox26, Thank you for submitting Square integration with Tendenci! We'll start to review it very soon.

jennyq:
Hi @theox26,

It works great with general payments!

However, I spotted two issues in recurring payments.

1) On memberships auto-renew, when a recurring payment is created, the `customer_profile_id` field is blank. The result is that the associated recurring payment doesn't have a payment method. (To replicate it,  enable recurring payments at /settings/module/recurring_payments/#id_enabled, turn on membership auto renewal /settings/module/memberships/#id_autorenew, then add a membership and select the checkbox auto renew.). 
image


2) When trying to add/update a payment method on recurring payment view page, for example https://yourdomain/rp/2/, an error occurred "Error: $.browser is undefined".  
image

Can you please look into those two issues? Let me know if further clarification is needed. Thanks!

theox26:
@jennyq I'll check these out. Our site has some customizations that might be why this wasn't an issue for us. I'll do a base install with this to test and let you know if I have any questions.

Feel free to check out www.keystonekidspace.org. It uses tendenci but is very different from the normal use case.

PR #1105, opened 2022-04-13
tag to pull all photos from one album

theox26:
A tag to allow pulling just images from a single album. This allows better use of the photo album app throughout the site.

Here is an example image carousel:

`<div id="img-carousel" class="carousel slide" data-ride="carousel">
      <div class="carousel-inner" role="listbox">
            {% album_photos 2 as photos %}
            {% for photo in photos %}
            <div class="item{% if forloop.first %} active{% endif %}">
              <div class="carousel-img" style="background-image:url('{% url "photo.size" photo.pk '500x400' %}')"></div>
            </div>
            {% endfor %}
      </div>
</div>`

It seems to remove most of the example but it still gives the django template part.

### DIFF
--- a/tendenci/apps/photos/templatetags/photo_tags.py
+++ b/tendenci/apps/photos/templatetags/photo_tags.py
@@ -105,6 +105,43 @@ def public_photos(parser, token, use_pool=False):
 def public_pool_photos(parser, token):
     return public_photos(parser, token, use_pool=True)
 
+class AlbumPhotosNode(Node):
+    def __init__(self, context_var, album_var=None):
+        self.context_var = context_var
+        if album_var is not None:
+            self.album_var = Variable(album_var)
+        else:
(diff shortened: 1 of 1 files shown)


jennyq:
Hi @theox26,

Thank you for submitting the pull request! However, this feature can be achieved with `list_photo_sets`.

    {% list_photo_sets as photo_sets_list filters="id=1" %}
        <ul>
          {% for photo_set in photo_sets_list %}
              {% for photo in photo_set.get_images %}
            <div class="carousel-img" style="background-image:url('{% url  "photo.size" photo.pk '500x400' %}')">
              {% endfor %}
          {% endfor %}
        </ul>

I've just added an example to the Doc. You can check it out here: 
https://tendenci.readthedocs.io/en/latest/templating/theme_code_snippets.html#photos-feed.

Issue #1109, opened 2022-06-08
RSS feed not up to date


bernd-wechner:
I can't find any lucid docs on the RSS feed, and it looks like a code drill down will be needed. But on our site we want the posted news items to appear as an RSS feeed. And they do. But not with rhyme or reason. We cannot understand the timing etc. Right now for example we have the second last posted news item on the RSS feed, and that last posted news but not the last one which was posted 3 days ago ... both are identically configured for RSS:

![image](https://user-images.githubusercontent.com/7296506/172596674-125b4764-c278-4a92-ab9b-086785c4552c.png)

And the second last one is in the feed but not the last one. 

This can be seen right now (June 8, 20220) by comparing:

https://hogs.org.au/news/
https://hogs.org.au/rss/

To see the latter you'll want an RSS reader or an XML formatter. The head of the later oddly says right now:

```
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0"
	xmlns:atom="http://www.w3.org/2005/Atom">
	<channel>
		<title>HoGS RSS Feed</title>
		<link>https://hogs.org.au/rss</link>
		<description>All syndicated rss feeds on HoGS</description>
		<atom:link href="https://hogs.org.au/rss/" rel="self"></atom:link>
		<language>en-us</language>
		<lastBuildDate>Tue, 31 May 2022 20:19:00 +1000</lastBuildDate>
```
after which items appear.

The last item in news has:

![image](https://user-images.githubusercontent.com/7296506/172597487-bbbe224e-ec93-44d6-91f6-d6f282301d3f.png)

while the second last one has:

![image](https://user-images.githubusercontent.com/7296506/172597633-03910f55-03ee-4927-a206-0a31f19b54b9.png)

And therein lies a big clue. Seems the second last one was released before the last build date.

Which raises the question why is the last build a week ago and can we schedule a nightly RSS build or a frequency of our choice? Can't find docs on it alas nor can I find it in site settings either through the web interface or the settings.py of the site or Tendenci's etc.

bernd-wechner:
A quick experiment, dumb blind one, I did just before I retired was to set MAX_FEED_ITEMS_PER_APP much higher and reload uwsgi (the site). And the latest news item now appears. 

I have a feeling that we have a bug here. 

[...]

PR #1110, opened 2022-06-12
App specific RSS requests

bernd-wechner:
The Main RSS feed is noisy with lots of apps contributing.

This small change implements app specific feeds. 

It was developed while investigating the cause of:

https://github.com/tendenci/tendenci/issues/1109

The cause has I believe been identified. Tendenci keeps an RSS cache in `self.all_items` so once requested once, new news can be written and not appear in the RSS feed until this cache is cleared. It seems to have the lifespan of the process serving the feed to me. Which is why it comes good when the uwsgi server is restarted.

The solution to that is not obvious and feedback will be sought. Solutions can include:

1. The rss app checking the latest object in each app and invalidating the cache is a new one is found. Sort of diminishes the benefit of caching.

2. The cache can simply be disabled, as an option in global settings for sites that don't want this behaviour.

3. Every app has the responsibility when saving new objects of invalidating the RSS cache. This can conceivably be done by globally overrriding the object save methods but might be overkill too.

Issues encountered in this change:

1. Class renamed from GlobalFeed to to RSSFeed
2. app specific caches also kept
3. the reverse URL is still `rss.mainrssfeed` didn't think a change prudent as sites may use this in their templates. 
4.`feedsmanager` now has `get_app_feed` to complement `get_all_feeds` which uses import lib to get the LatestEntriesFeed class
5. The site wide URLS support an optional app prefix in the rss app URL (so the URL is an RSS feed on a given app).


bernd-wechner:
Sorry I've become so distracted of late, and dropped the ball on this. I shall have to make a slot for Tendenci stuff some time soon.  I'll review this some time too. I would like to see these URLs as they are slightly more intuitive this way, but the /feed ones, and on a central root. But will come round to a full review in time of my various PRs and their status. Thanks heaps Jenny!

PR #1114, opened 2022-07-10
Form memory model

bernd-wechner:
As promised, an update to form memory management with:

1. A new model to store the form memories
2. A new view to inspect them
3. Updates to menus to show this new form (I can find the `top_nav_itmes.html` in the UI to test, but have not found where `nav.html` and `options.html` are used to test)
4. Updates to form handling to write to the model and read from it.
5. A management command "migrate_memoies" that will populate the model from legacy Session store memories. Can be run once to migrate the session based memories.
6. A new template tag `startswith` in the forms app, as I needed one to display the memories well.

Hope this passes muster and can merge cleanly. Alas, a few code changes to review here. Most of the 'messy' changes relate simply to dropping `gfield_key` as a session store key, and using the new model. This model can now be efficiently queried for the memories view and permits future improvements to manage memories. Delete them for example (one shortfall of the current implementation - to fix - is that Anonymous session memories don't expire). One possible fix here, is simply to set on_delete to CASCADE and guaranteed that user keyed memories None for a session so they are not inadvertently deleted when an associate session expires.

Let me kn ow if you'd like that implemented up front. Parking this as a satisfactory release candidate for not. After some merging mess because by dev branch is diverging because https://github.com/tendenci/tendenci/pull/1062 is not merged yet ... and there is overlapping code.


bernd-wechner:
A fairly heavy review and test session on this. It works well

bernd-wechner:
Appreciate feedback and megring before brnaches converge too much. And any commentary. Am particularly keen to know if the migration file merges well.

tendenci/apps/forms_builder/forms/urls.py
```diff
@@ -10,25 +10,34 @@
...
-    re_path(r"^%s/entry/(?P<id>\d+)$" % urlpath, views.entry_detail, name="form_entry_detail"),
-    re_path(r"^%s/entries/(?P<id>\d+)$" % urlpath, views.entries, name="form_entries"),
     re_path(r"^%s/payment/(?P<invoice_id>\d+)/(?P<invoice_guid>[\d\w-]+)?/$" % urlpath, views.form_entry_payment, name="form_entry_payment"),
 
+    # Form use
     re_path(r"^%s/files/(?P<id>\d+)$" % urlpath, views.files, name="form_files"),
-    re_path(r"^%s/(?P<slug>.*)/sent/$" % urlpath, views.form_sent, name="form_sent"),
-    re_path(r"^%s/(?P<slug>.*)/$" % urlpath, views.form_detail, name="form_detail"),
+    re_path(r"^%s/(?P<slug>[\w-]+)/sent/$" % urlpath, views.form_sent, name="form_sent"),
+    re_path(r"^%s/(?P<slug>[\w-]+)/$" % urlpath, views.form_detail, name="form_detail"),
```
jennyq, line 28: The url pattern for the "form_detail" needs to remain at the bottom of the list because the "slug" pattern can match any url underneath it.

bernd-wechner: Happy to do that, but keen to understand. Underneath it in that same urlpatterns block I assume, and I can't see how it would match any beneath it. 

```py
"^%s/(?P<slug>[\w-]+)/$"
```

presumably becomes effectively:

```py
"^https://tendenci-site.org/forms/(?P<slug>[\w-]+)/$"
```
presuming that `urlpath` is `https://tendenci-site.org/forms` at the time of request routing.

and beneath it are:

```py
    re_path(r"^%s/entries/(?P<id>\d+)$" % urlpath, views.entries, name="form_entries"),
    re_path(r"^%s/entries/delete/(?P<id>\d+)$" % urlpath, views.entry_delete, name="form_entry_delete"),
    re_path(r"^%s/entries/export/(?P<id>\d+)/full$" % urlpath, views.entries_export,
            {'include_files': True, }, name="form_entries_export_full"),
    re_path(r'^%s/entries/export/(?P<task_id>[-\w]+)/status/$' % urlpath, views.entries_export_status, name="form_entries_export_status"),
    re_path(r'^%s/entries/export/(?P<task_id>[-\w]+)/check/$' % urlpath, views.entries_export_check, name="form_entries_export_check"),
    re_path(r'^%s/entries/export/(?P<task_id>[-\w]+)/download/$' % urlpath, views.entries_export_download, name="form_entries_export_download"),

    re_path(r"^%s/entry/(?P<id>\d+)$" % urlpath, views.entry_detail, name="form_entry_detail"),

    # Form memory management
    re_path(r"^%s/memories/(?P<id>\d+)/?$" % urlpath, views.memories, name="form_memories")
```

none of which I can matching that pattern. My reading is based on the trailing $ the zero width end of string match.

I'm an avid learner and keen to understand.

jennyq: You're right. I didn't notice you've changed the slug pattern from (?P<slug>.*) to (?P<slug>[\w-]+), which prevents it from matching any of the patterns underneath it. The dot (.) matches any character except a newline.  This change is good because the actual pattern of slug in forms is '^[-a-zA-Z0-9_]+\Z'.

bernd-wechner: Thanks, Jenny. \Z is new to me, and interesting, like $ but more extreme.  Can URLs span more than one line though? If not $ and \Z collapse to the same. Just curious really.

\w though is the same as [-a-zA-Z0-9_] (though it leaves implementation flexibility matching against Unicode characters like ä or è etc. - which depends entirely on the Python interpreter's low level implementation of it)

I admit I took the form I used from here:

https://stackoverflow.com/questions/23358685/regular-expression-in-url-for-django-slug

As I failed to find a doc anywhere in the Django references as to the legal character set for slugs.
Finally: a quick side question. f-strings of course are the new norm since Python 3.6 and far more legible, so on most projects I'm slowly as I encounter them moving % notations to f-strings. Is that something the tendenci project supports? In this instance I'd move these re_path args from: `r"^%s/entries/(?P<id>\d+)$" % urlpath` to `fr"^{urlpath}/entries/(?P<id>\d+)$"` for example.

jennyq: RE "Can URLs span more than one line though?",  likely can, but I'm not sure. Though it doesn't make sense to me to support link breaks in URLs. 

As to the django's slug field, if `allow_unicode` is set to True, it uses the pattern `r"^[-\w]+\Z"` instead of `r"^[-a-zA-Z0-9_]+\Z` https://github.com/django/django/blob/main/django/core/validators.py#L260. The slug field in Tendenci forms module doesn't set `allow_unicode`, so we're using the ASCII character set. The django doc has most of the info except the specific character sets https://docs.djangoproject.com/en/3.2/ref/models/fields/#slugfield 

Currently, Tendenci supports Python 3.7 and newer, so it's okay to use f-strings in url patterns.

bernd-wechner: Voilà, looks real nice now. I opted to try and set a clear standard for slug REs ;-). In fact, methinks this could be tendenci wide, and the slug_re be defined in an application wide configuration somewhere too. I was keen to use django.core.validators.slug_re, but can't because the URL patterns match a slug in the middle of  a longer string (a URL) and the validators definition includes a ^ and \Z bounds check (not appropriate in the URL context). 

I take the view that such standard REs (and I would include uuid, and object ID matches in that as standard terms centrally  defined. Makes the URL patterns easier to read, and avoids individual patterns or apps diverging from norms.

bernd-wechner:
No such luck. Looks to be a significant Github bug. At the top of this PR it states:

[bernd-wechner](https://github.com/bernd-wechner) wants to merge 395 commits into [tendenci:master](https://github.com/tendenci/tendenci/tree/master) from [bernd-wechner:form-memory-model](https://github.com/bernd-wechner/tendenci/tree/form-memory-model)

And the link to bernd-wechner:form-memory-model takes us here:

https://github.com/bernd-wechner/tendenci/tree/form-memory-model

where it says clearly at top:

This branch is [1 commit ahead of](https://github.com/bernd-wechner/tendenci/compare/tendenci%3Atendenci%3Amain...form-memory-model) tendenci/tendenci:main.

Also on my branch the Commit Button does not offer me the option to push the branch upstream but only to open a new PR. And so we must conclude a github bug (inconsistent state) and the best way forward is to close this and issue a new PR as the github UI offers.

PR #1142, opened 2022-11-28
Explicitly specify the encoding when loading theme templates

BenSturmfels:
After migrating a Tendenci install to a new Ubuntu 22.04 server, I found we were getting errors like "UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 13650: ordinal not in range(128)" (full backtrace included in GitHub comment). This was resolved by adding an explicit encoding to the calls to open.

I wasn't aware, but specifying the `encoding` when opening text files that were not necessarily created in the current locale appears to be a best practice. Details here: https://peps.python.org/pep-0597/ and https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unspecified-encoding.html. I believe that this advice applies here because some theme files may have been created locally, but others may come from upstream.

### DIFF
--- a/tendenci/apps/theme/template_loaders.py
+++ b/tendenci/apps/theme/template_loaders.py
@@ -83,7 +83,7 @@ class ThemeLoader(DjangoLoader):
     def get_contents(self, origin):
         if not origin.use_s3_theme:
             try:
-                with open(origin.name) as fp:
+                with open(origin.name, encoding='utf-8') as fp:
                     return fp.read()
             # Python 3 only
             #except FileNotFoundError:
--- a/tendenci/apps/theme_editor/utils.py
+++ b/tendenci/apps/theme_editor/utils.py
@@ -242,7 +242,7 @@ def get_file_content(root_dir, theme, filename):
     if not content:
         current_file = os.path.join(root_dir, filename)
         if os.path.isfile(current_file):
-            fd = open(current_file, 'r')
+            fd = open(current_file, 'r', encoding='utf-8')
             content = fd.read()
             fd.close()
     return content


BenSturmfels:
Here's the backtrace:

```
/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/tendenci/themes/t7-base/templates/boxes/edit-link.html 2022-11-20 11:20:53 ERROR django.request: Internal Server Error: /dashboard/ Traceback (most recent call last):
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/tendenci/apps/dashboard/views.py", line 111, in new
    return render_to_resp(request=request, template_name=template_name, context={
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/tendenci/apps/theme/shortcuts.py", line 38, in themed_response
    template = get_template(template_name, using=using)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/template/loader.py", line 15, in get_template
    return engine.get_template(template_name)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/template/backends/django.py", line 34, in get_template
    return Template(self.engine.get_template(template_name), self)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/template/engine.py", line 143, in get_template
    template, origin = self.find_template(template_name)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/template/engine.py", line 125, in find_template
    template = loader.get_template(name, skip=skip)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/django/template/loaders/base.py", line 24, in get_template
    contents = self.get_contents(origin)
  File "/srv/venvs/xxxx-django-cp310/lib/python3.10/site-packages/tendenci/apps/theme/template_loaders.py", line 88, in get_contents
    return fp.read()
  File "/usr/lib/python3.10/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 13650: ordinal not in range(128)
```

BenSturmfels:
Thanks Jenny. Interestingly I'm getting "utf-8" for both:

```
$ /srv/venvs/xxxx-django-cp310/bin/python manage.py shell -c "import sys; print(sys.getdefaultencoding())"
utf-8
$ locale
LANG=C.UTF-8
LANGUAGE=
LC_CTYPE="C.UTF-8"
LC_NUMERIC="C.UTF-8"
LC_TIME="C.UTF-8"
LC_COLLATE="C.UTF-8"
LC_MONETARY="C.UTF-8"
LC_MESSAGES="C.UTF-8"
LC_PAPER="C.UTF-8"
LC_NAME="C.UTF-8"
LC_ADDRESS="C.UTF-8"
LC_TELEPHONE="C.UTF-8"
LC_MEASUREMENT="C.UTF-8"
LC_IDENTIFICATION="C.UTF-8"
LC_ALL=
```

I'm going to do some more digging to see if I can find out why it's not using utf-8. It's definitely still an issue for us - I just tried without the patch again and am getting same error. Prior to the server move we'd be stuck on Tendenci ~ 12 due to having an older Python, so I'm not 100% sure that it's not server related, or a combination of the server AND the Tendenci upgrade.

BenSturmfels:
I've got to the bottom of this - was a server configuration issue.

The issue was that in our uWSGI config I had specified `env = LANG=en_AU.UTF-8`, but it turns out that this encoding wasn't installed on the system, so when `locale.getpreferredencoding(False)` was called by `open()`, it was getting back `ANSI_X3.4-1968`. The `fp.read()` was then choking on a curly quote in our `homepage.html` theme template. The challenge in hunting this down was that the code was behaving differently in the Python interpreter directly because it didn't have the custom `LANG` set!

That custom `LANG` had become part of my standard deployment recipe to avoid errors when people upload files with UTF-8 characters in the filenames. Going forward though, I'm instead moving to setting `PYTHONUTF8=1`. This forces Python to use UTF-8 for reading and writing rather than the system locale and will become the default in Python 3.15.

https://docs.python.org/3.8/library/functions.html#open
https://docs.python.org/3/library/locale.html
https://docs.python.org/3/library/os.html#utf8-mode

BenSturmfels:
Oh and interesting, uWSGI Emperor wasn't clearing the `LANG` environment variable when reloaded by `touch`ing the config file as I would have expected. It took a full restart to clear it.

PR #1160, opened 2023-03-03
Add CEU table in events

shandy-bixly:

PR #1170, opened 2023-05-17
Add parent child events

shandy-bixly:


shandy-bixly:
I needed code from PR #1160 so this branch is based off shandy-bixly:add-credits-table & needs to be merged after #1160

jennyq:
An error occurred when running `python manage.py update_dashboard_stats`.
`django.db.utils.IntegrityError: Problem installing fixture '.../tendenci/apps/events/fixtures/tendenci_default_events.json': Could not load events.RegistrationConfiguration(pk=1): null value in column "cancellation_fee" of relation "events_registrationconfiguration" violates not-null constraint
`

The migration "0021_auto_20230511_1724" somehow added some fields, for example, "cancellation_fee", that do not exist in this branch. Anyway, some migrations will need to be rebuilt after merged. So, you can disregard it for now.

jennyq:
1. On event add/edit, please move the "Event code" field down below the "Short name", keep the Title at the top. Also add some help text for "Event code" and "Short name" fields as what they're for.
2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse? For a parent event, please add a link to add a sub event. When the link is clicked, auto-populate the fields "Parent" and "Event relationship". For a child event, add a link to the parent event.

tendenci/themes/t7-base/templates/events/reg8n/include/register-registrant.js
```diff
@@ -68,6 +71,17 @@ function updateFormHeader(this_form, prefix, idx){
 
 };
 
+function updateAddRegistrantButton(formCount) {
+    if (guest_limit == 0) { return; }
```
jennyq, line 75: The `guest_limit` is undefined. It should be moved right beneath it is defined.

shandy-bixly:
> 1. On event add/edit, please move the "Event code" field down below the "Short name", keep the Title at the top. Also add some help text for "Event code" and "Short name" fields as what they're for.
> 2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse? For a parent event, please add a link to add a sub event. When the link is clicked, auto-populate the fields "Parent" and "Event relationship". For a child event, add a link to the parent event.

1. For the Short Name, I know this is for the

shandy-bixly:
> 2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse? 

Adding Child to Parent to the menu will be easy. However, I'm not sure about Parent to Child since there are multiple Child Events (could result in a very long menu) Also, if it's a Parent, it already has a list of the Child Events both in edit & view mode.

jennyq:
> > 2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse?
> 
> Adding Child to Parent to the menu will be easy. However, I'm not sure about Parent to Child since there are multiple Child Events (could result in a very long menu) Also, if it's a Parent, it already has a list of the Child Events both in edit & view mode.

The link from Parent would be a link to **add** a child (sub) event. No need to link to each child from top menu.

shandy-bixly:
> > 2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse?
> 
> Adding Child to Parent to the menu will be easy. However, I'm not sure about Parent to Child since there are multiple Child Events (could result in a very long menu) Also, if it's a Parent, it already has a list of the Child Events both in edit & view mode.

All other changes requested have been made.  I think we should discuss adding the child events to the menu before doing that as it could result in a very long menu.... going to go ahead & push up changes and re-request review.
@jennyq  @ndavis2

shandy-bixly:
> > > 2. On the event top menu, under the left section "This Event", can you please add some links for easy navigation from a parent to a child, and vice verse?
> > 
> > 
> > Adding Child to Parent to the menu will be easy. However, I'm not sure about Parent to Child since there are multiple Child Events (could result in a very long menu) Also, if it's a Parent, it already has a list of the Child Events both in edit & view mode.
> 
> The link from Parent would be a link to **add** a child (sub) event. No need to link to each child from top menu.

ok, will add the add link

jennyq (approved):

PR #1173, opened 2023-05-30
Create event badge

shandy-bixly:
Only the last 2 commits are actually for this PR. I will clean this up once the proceeding PRs are merged. (I needed the parent/child event code in order to print child events to the back of the badge).


shandy-bixly:
Only the last 2 commits are actually for this PR. I will clean this up once the proceeding PRs are merged. (I needed the parent/child event code in order to print child events to the back of the badge).

tendenci/themes/t7-base/templates/events/badge.html
```diff
@@ -0,0 +1,165 @@
...
+          <div>
+            {{ registrant.company_name }}
+          </div>
+          <div>
+            {{ registrant.pricing.title }}
+          </div>
+        </div>
+        <div class="row">
+          <div class="col-md-6 qr">
+            {% qr_from_text name size="s" %}
```
jennyq, line 128: Hi Shandy,
The QR Code generated from `{% qr_from_text name size="s" %}` are all identical because the variable "name" is not defined. Based on Nicole's notes, the QR Code should be unique and linked to member/user’s registration.

shandy-bixly: @jennyq This is a placeholder that does nothing. I think I was originally passing in the name... but changed something so it no longer does that. 
The reason, I'm using a placeholder is that the link it will point to doesn't exist yet. This is also why I added the USE_BADGES setting... so we can turn it off for everyone by default.  

The link will be added in the digital check in story. Ideally, we would have worked on that feature before this, but I got started on badges because we didn't have check in defined at that time.  (we have it defined now, but registration has to be complete before we can get started on digital check in)

Another piece that's coming soon is that currently, the back of the badge displays all child events tied to an event. In the 'add registration' work that I'm currently working on, it will be updated to only display the ones the registrant is registered for... but the full list is used for now as a placeholder.

jennyq: Gotcha. Thanks for explaining it. Since the badge code is already in the 'add-registration' branch, we can review it along with the registration update.

PR #1181, opened 2023-07-10
Add registration

shandy-bixly:


jennyq:
Shandy,

Noticed two issues. You might be able to quick fix them:
1) Some css code on the sub-events registration breaks the format of the Events tab on top menu.
image

2) IndexError at /events/registration/11/edit/. Thought it worked when you demoed last week. Maybe the change is not in this branch?

      ```
    File ".... tendenci/apps/events/views.py", line 2978, in registration_edit
      len(updated_attendance_dates[index]) + past_dates
     IndexError: list index out of range
    "POST /events/registration/13/edit/ HTTP/1.1" 500 139528
    ```

image

jennyq:
3) Also on registration edit, if the setting "Enable Nested Events" is turned OFF, it throws TypeError: 'NoneType' object is not iterable

       File "... /tendenci/apps/events/models.py", line 1191, in past_attendance_dates
       return [x for x in self.attendance_dates if parse(x).date() <= datetime.now().date()]
       TypeError: 'NoneType' object is not iterable

jennyq:
4) On event pricing edit, can we hide the "Days price covers" field if the setting "Enable Nested Events" is not turned on?

shandy-bixly:
@jennyq 
I'm not able to repro the Event menu bug:
image

shandy-bixly:
> 2\. IndexError at /events/registration/11/edit/. Thought it worked when you demoed last week. Maybe the change is not in this branch?
>     ```
>     File ".... tendenci/apps/events/views.py", line 2978, in registration_edit
>     len(updated_attendance_dates[index]) + past_dates
>     IndexError: list index out of range
>     "POST /events/registration/13/edit/ HTTP/1.1" 500 139528
>     ```
> 
> image

That 2nd one is happening only if all dates are in the past. In that case, they shouldn't be able to change the dates because they would have already attended & we don't want them getting 3 extra dates for free. However,  this works if some dates are in the past but others are still upcoming.... if there's no upcoming dates then we get the IndexError... this will be an easy fix though

jennyq:
> @jennyq I'm not able to repro the Event menu bug: image

It's on sub-event registration, **not** on event view.

jennyq:
> That 2nd one is happening only if all dates are in the past. In that case, they shouldn't be able to change the dates because they would have already attended & we don't want them getting 3 extra dates for free. However, this works if some dates are in the past but others are still upcoming.... if there's no upcoming dates then we get the IndexError... this will be an easy fix though @jennyq

Actually, the one I tested, none of the sub-events are in the past. They're from 9/2/2023 to 11/17/2023.

shandy-bixly:
> > That 2nd one is happening only if all dates are in the past. In that case, they shouldn't be able to change the dates because they would have already attended & we don't want them getting 3 extra dates for free. However, this works if some dates are in the past but others are still upcoming.... if there's no upcoming dates then we get the IndexError... this will be an easy fix though @jennyq
> 
> Actually, the one I tested, none of the sub-events are in the past. They're from 9/2/2023 to 11/17/2023.

Interesting... the only way I was able to repro it was if all the dates were in the past.

shandy-bixly:
> > That 2nd one is happening only if all dates are in the past. In that case, they shouldn't be able to change the dates because they would have already attended & we don't want them getting 3 extra dates for free. However, this works if some dates are in the past but others are still upcoming.... if there's no upcoming dates then we get the IndexError... this will be an easy fix though @jennyq
> 
> Actually, the one I tested, none of the sub-events are in the past. They're from 9/2/2023 to 11/17/2023.

I'm unable to reproduce the bug on my environment. Can you give me some details on the setup?
1. What is the date range for the parent event (as well as the child events registered for)
2. What does registrant.attendance_dates have?

jennyq:
The error "IndexError: list index out of range" is gone now after I pulled down your updates. Looks like that exception has been caught. Now it shows an alert message "Select 1 dates for ..."

image
I was testing a migrated event with 55 pricings yesterday. Since that one is too complex to show here, I created a simple one. The event is from 9/1/2023 to 9/3/2023. 3 sub-events are on 9/1, 9/2, 9/3, respectively. The "Number of attendees"  for both pricings is set to 1. 

image

Let me know if you need additional info. 

Thanks

shandy-bixly:
> The error "IndexError: list index out of range" is gone now after I pulled down your updates. Looks like that exception has been caught. Now it shows an alert message "Select 1 dates for ..."
> 
> image I was testing a migrated event with 55 pricings yesterday. Since that one is too complex to show here, I created a simple one. The event is from 9/1/2023 to 9/3/2023. 3 sub-events are on 9/1, 9/2, 9/3, respectively. The "Number of attendees" for both pricings is set to 1.
> 
> image
> 
> Let me know if you need additional info.
> 
> Thanks

What is `registrant.attendance_dates` for Anna Smith?

jennyq:
> What is `registrant.attendance_dates` for Anna Smith? 

The value of the attendance_dates is ` ["2023-09-01"]` in the table "events_registrant".

shandy-bixly:
> > What is `registrant.attendance_dates` for Anna Smith?
> 
> The value of the attendance_dates is ` ["2023-09-01"]` in the table "events_registrant".

ok thanks. I will do a similar setup on my computer & see if I can duplicate/fix the error

shandy-bixly:
> ["2023-09-01"]

@jennyq 

Sadly I'm still unable to duplicate the error. I've tried several things & it is always working for me. The only scenario I found was the one that I mentioned above (if all dates were older). There is a fix to the attendance_dates in the check_in branch that I found when I was experimenting with a past event in July due to formatting differences... (July wasn't abbreviated so it didn't match the formatting I was  using to try to get the option included). However, I don't think that is causing this since September is also abbreviated... also I haven't had any issues testing with September... 

This is the line that is fixed in the check in branch... but I don't think it's related to the bug you're seeing here (formatted_attendance_dates = [datetime.strptime(x, '%b. %d, %Y').strftime("%Y-%m-%d") for x in attendance_dates])

shandy-bixly:
@jennyq  

For it to get that specific error... something has to be failing this if:
if pricing and pricing.days_price_covers and total_attendance_dates != pricing.days_price_covers:

shandy-bixly:
@jennyq 

I pushed up a fix for the attendance dates formatting. I don't think this actually addresses the bug you were getting, but it was the only thing I could find. I couldn't duplicate the issue on my end. The only way I could get the error from your screen shot is by selecting 2 dates instead of just 1. 

The only way you get that error is if this statement is False:
 past_dates = len(registrant.past_attendance_dates)
total_attendance_dates = len(registrants_updated_dates) + past_dates
...
if pricing and pricing.days_price_covers and total_attendance_dates != pricing.days_price_covers:

Can you verify that `days_price_covers` is set to 1?
Also,  What is registrant.past_attendance_dates? I'm thinking in the scenario you gave it should be 0

jennyq:
Shandy, I think I've figured out why you couldn't replicate the issue. When you testing, were you using a custom reg form? This bug is in the regular registration form.
https://github.com/shandy-bixly/tendenci/blob/add-registration/tendenci/apps/events/views.py#L2911

shandy-bixly:
> Shandy, I think I've figured out why you couldn't replicate the issue. When you testing, were you using a custom reg form? This bug is in the regular registration form. https://github.com/shandy-bixly/tendenci/blob/add-registration/tendenci/apps/events/views.py#L2911

@jennyq  That was exactly it! I think I have it working now. Just pushed up a fix

PR #1183, opened 2023-07-25
Use TZ datetimes with overrided `date` filter

DylannCordel:
13 years ago, django `date` filter has been overrided in tendency via commit @47e0914e

It was compatible with django date. But 1 year later, django adds support for timezone with the  `date` filter and this change (commit django/django@9b1cb755a28f020e27d4268c214b25315d4de42e) has not been backported in tendenci, resulting all dates displayed in templates with `date` filter are displayed in UTC, whatever the current (user|site) timezone is.

### DIFF
--- a/tendenci/apps/base/templatetags/base_filters.py
+++ b/tendenci/apps/base/templatetags/base_filters.py
@@ -80,7 +80,7 @@ def date_long(value, arg=None):
             return ''
 date_long.is_safe = False
 
-@register.filter_function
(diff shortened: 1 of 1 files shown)


jennyq:
Nice fix! Thanks @DylannCordel. 

With Postgresql, we don't have the need to turn on the `USE_TZ` setting. That's why this issue hadn't been noticed for soooo long.

PR #1190, opened 2023-07-31
Digital check in

shandy-bixly:


tendenci/apps/events/views.py
```diff
@@ -3871,6 +3872,55 @@ def registrant_roster(request, event_id=0, roster_view='', template_name='events
         'total_checked_in': total_checked_in})
 
 
+@is_enabled('events')
+@login_required
+def digital_check_in(request, registrant_id, template_name='events/reg8n/checkin.html'):
```
jennyq, line 3877: Can we add some permission check for digital_check_in? Right now, looks like all logged-in users can check in and check out anybody.

    if not has_perm(request.user, 'events.view_registrant'):
        raise Http403

shandy-bixly: yes, looks like I missed that... will add it & push the code back up

shandy-bixly: updated

jennyq (approved):

jennyq:
Shandy, 

After it is merged, I noticed an issue: the badge image doesn't show in the confirmation email in Gmail (not sure if it works with other email clients). Looks like Gmail doesn't support SVG images, possibly for the security reason.

I've tried `{% qr_from_text registrant.check_in_url size="T" image_format="png" %}` to render as a PNG image, but it doesn't work either... looks like base64 images are not supported by Gmail as well. 

The img tag with `qr_url_from_text` works. This might add a bit overhead though. 
`Badge`

I'll push out the update. Just an FYI. Thanks.

shandy-bixly:
ok, thanks. I will do some research to see if there's a better way... if I
don't find anything, I'll use the URL

PR #1192, opened 2023-08-03
use django-admin executable to call commands

zerrac:
`subprocess.Popen([python_executable(), "manage.py", ...]` wont work if manage.py is not directly in the path of the subprocess.

Plus the use of django-admin remove the need of a complex python executable selector

### DIFF
--- a/tendenci/apps/articles/views.py
+++ b/tendenci/apps/articles/views.py
@@ -426,7 +426,7 @@ def export(request, template_name="articles/export.html"):
         default_storage.save(temp_file_path, ContentFile(b''))
 
         # start the process
-        subprocess.Popen([python_executable(), "manage.py",
+        subprocess.Popen(["django-admin",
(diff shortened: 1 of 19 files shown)


jennyq:
Hi @zerrac,

Nice concept! However, it didn't work when I tested it.  Looks like the `django-admin`  works ONLY with django's builtin management commands. To make it work with the commands in Tendenci, some environment variables would need to be set up.  
1) export PYTHONPATH=(site root)
2) export DJANGO_SETTINGS_MODULE=conf.settings 
3) export TENDENCI_PROJECT_ROOT=(site root) 

With that being said, I'm not sure if it is still a good idea to replace python_executable(), "manage.py" with django-admin.

jennyq:
We've never run into the need to "configure the path to manage.py". How did you set up your tendenci site differently?

PR #1196, opened 2023-08-07
Ceu override

shandy-bixly:
Backend models & logic for CEU Overrides. Creating this PR separately so that migrations can continue even while certificate layout is being worked on. (certificate layout won't require any tables).


jennyq:
Shandy,

As we talked about it in the meeting today, here are two fields that need to be added to the RegistrantCredits model for the performance reason. Thanks.

    event = models.ForeignKey('Event', on_delete=models.CASCADE)
    credit_dt = models.DateTimeField(db_index=True)

shandy-bixly:
> Shandy,
> 
> As we talked about it in the meeting today, here are two fields that need to be added to the RegistrantCredits model for the performance reason. Thanks.
> 
> ```
> event = models.ForeignKey('Event', on_delete=models.CASCADE)
> credit_dt = models.DateTimeField(db_index=True)
> ```

Thanks, going to add these & get it pushed up in just a few minutes

shandy-bixly:
> Shandy,
> 
> As we talked about it in the meeting today, here are two fields that need to be added to the RegistrantCredits model for the performance reason. Thanks.
> 
> ```
> event = models.ForeignKey('Event', on_delete=models.CASCADE)
> credit_dt = models.DateTimeField(db_index=True)
> ```

@jennyq updated & pushed code back up

PR #1201, opened 2023-08-16
Fix: Invoice 'Download PDF' asks to sign-in

Tilix4:
When not signed-in, the invoice button `Download PDF` asks to sign-in, which is confusing. If this operation isn't available for a anonymous person, I propose to hide it.


jennyq:
Thanks @Tilix4 ! Closing it as the issue has been resolved.

PR #1203, opened 2023-08-16
Add Credits Grid with links to Certificate to Profile

shandy-bixly:
@jennyq Separating this from certificates in case you want to merge the certificates first. There's a SignatureImage & CertificateImage model in the certificate branch... so we might want those in sooner. The logic for this one is ready, but I'm still adding styling. 

Feel free to comment on anything that's here already aside from styling


shandy-bixly:
@jennyq got the sytling applied. I'm still doing some testing to make sure longer names, etc don't break it, but I'm going to go ahead and remove the WIP

PR #1205, opened 2023-08-21
Include all dates on certificate

shandy-bixly:
Per Nicole: "Confirmed symposium cert dates are event dates, not just dates attended"

Updated to include all dates in events (not just the ones the registrant attended)
Since it is no longer registrant specific, was able to switch out the sample cert with all dates of event instead of current date as well.


shandy-bixly:
This PR includes all dates of the Event on the certificate rather than registrant.attendance_dates
This does not include filtering out Events outside of the parent Event's window... that will be in a separate PR. Once that separate PR is merged, the filtering will happen on the certificate as well

jennyq:
Shandy,

According to Nicole, for dates like "Nov 16, 2022, Nov 17, 2022, Nov 18, 2022", it should show "November 16, 17, 18, 2022".

I'm going to merge this PR. When you get the chance, can you tweak it please? Thanks!

PR #1207, opened 2023-08-22
Fix ceu grid

shandy-bixly:
Fixes sorting issue & missing columns when previous years have credits that later years don't.
Also sorts from most recent date 
Fixes formatting for details


jennyq:
Shandy,

Merged but noticed an issue. Can you please look at this format issue?
image

PR #1208, opened 2023-08-22
Register enabled child events only

shandy-bixly:


jennyq:
With this PR, I think the sub-events list should have additional column (for event editors) to show if registration is enabled or not so that event editors can easily see and adjust if needed. Otherwise, it'd be a quite burden for event editors, especially for events with many sub-events.

shandy-bixly:
> With this PR, I think the sub-events list should have additional column (for event editors) to show if registration is enabled or not so that event editors can easily see and adjust if needed. Otherwise, it'd be a quite burden for event editors, especially for events with many sub-events.

Do you mean on the Event edit where they see all the sub-events? Just making sure I know where to put it.  @jennyq

jennyq:
> Do you mean on the Event edit where they see all the sub-events? Just making sure I know where to put it. @jennyq

It'd be nice to show on both event edit and view. At the minimal, display it on event edit.

PR #1209, opened 2023-08-22
Exclude child events outside of Event window

shandy-bixly:


jennyq:
After this update, it also affects the sub-events list on event view and edit. If a sub-event, by any chance, is out of the range of the parent event, it would disappear from the sub-events list and no way to find it to edit. 

Is it possible to only handle it on registration but let them still show on parent event view and edit? Or, maybe it's better to not handle them at all. Let's discuss it with Nicole.

shandy-bixly:
> After this update, it also affects the sub-events list on event view and edit. If a sub-event, by any chance, is out of the range of the parent event, it would disappear from the sub-events list and no way to find it to edit.
> 
> Is it possible to only handle it on registration but let them still show on parent event view and edit? Or, maybe it's better to not handle them at all. Let's discuss it with Nicole.

updated

PR #1210, opened 2023-08-23
Allow admin register past child events

shandy-bixly:


jennyq:
Tested a past event registration as an admin. There is no Attendance dates list shows up on the registration form. On submit, it throws "Attendance dates" validation error.
image

shandy-bixly:
> Tested a past event registration as an admin. There is no Attendance dates list shows up on the registration form. On submit, it throws "Attendance dates" validation error. image

fixed

jennyq:
Still not able to register past child events as an admin.

shandy-bixly:
> Still not able to register past child events as an admin.

weird... it worked for me. I will take another look

shandy-bixly:
> > Still not able to register past child events as an admin.
> 
> weird... it worked for me. I will take another look

@jennyq  I looked over the code and it looks like while admins are allowed to see past child events, it is still filtering out ones that have registration_enabled=False and ones that have registration closed. Do we need to enable 1 or both of those options for admins?

jennyq:
> > > Still not able to register past child events as an admin.
> > 
> > 
> > weird... it worked for me. I will take another look
> 
> @jennyq I looked over the code and it looks like while admins are allowed to see past child events, it is still filtering out ones that have registration_enabled=False and ones that have registration closed. Do we need to enable 1 or both of those options for admins?

Just skip the check `if registrant.registration_closed` for admin in the view `register_child_events` (Ln 2118 in events/views.py). That should be enough. 

Also, please add some security check for the view `register_child_events`. (Otherwise, users' child event registration can be easily messed up by an attacker.) You can apply the same logic of how `cancel_registration` handles the perms check. Thanks!

shandy-bixly:
> > > > Still not able to register past child events as an admin.
> > > 
> > > 
> > > weird... it worked for me. I will take another look
> > 
> > 
> > @jennyq I looked over the code and it looks like while admins are allowed to see past child events, it is still filtering out ones that have registration_enabled=False and ones that have registration closed. Do we need to enable 1 or both of those options for admins?
> 
> Just skip the check `if registrant.registration_closed` for admin in the view `register_child_events` (Ln 2118 in events/views.py). That should be enough.
> 
> Also, please add some security check for the view `register_child_events`. (Otherwise, users' child event registration can be easily messed up by an attacker.) You can apply the same logic of how `cancel_registration` handles the perms check. Thanks!

sounds good.... about to make these updates

shandy-bixly:
> > > > > Still not able to register past child events as an admin.
> > > > 
> > > > 
> > > > weird... it worked for me. I will take another look
> > > 
> > > 
> > > @jennyq I looked over the code and it looks like while admins are allowed to see past child events, it is still filtering out ones that have registration_enabled=False and ones that have registration closed. Do we need to enable 1 or both of those options for admins?
> > 
> > 
> > Just skip the check `if registrant.registration_closed` for admin in the view `register_child_events` (Ln 2118 in events/views.py). That should be enough.
> > Also, please add some security check for the view `register_child_events`. (Otherwise, users' child event registration can be easily messed up by an attacker.) You can apply the same logic of how `cancel_registration` handles the perms check. Thanks!
> 
> sounds good.... about to make these updates

@jennyq  updated!

jennyq:
Thanks Shandy for the update!

One more issue:

The `hash` in the perms check needs to be passed in from url for anonymous users. Actually, we shouldn't name the variable `hash` as it is a Python built-in function. (Anyway, that is an issue in `cancel_registration`.) Let's use `guid` with another url pattern. 

    re_path(r'^%s/register/(?P<registration_id>\d+)/sub-events/(?P<guid>[\d\w-]+)/$' % urlpath, views.register_child_events, name='event.register_child_events'),

Then, in views.py,

    def register_child_events(request, registration_id, guid=None,  template_name="events/reg8n/register_child_events.html"):
        registration = get_object_or_404(Registration, pk=registration_id)

        perms = (
            has_perm(request.user, 'events.change_registration', registration),  # has perm
            request.user.is_authenticated and request.user == registration.registrant.user,  # main registrant
            registration.guid == guid, 
        )

shandy-bixly:
> Thanks Shandy for the update!
> 
> One more issue:
> 
> The `hash` in the perms check needs to be passed in from url for anonymous users. Actually, we shouldn't name the variable `hash` as it is a Python built-in function. (Anyway, that is an issue in `cancel_registration`.) Let's use `guid` with another url pattern.
> 
> ```
> re_path(r'^%s/register/(?P<registration_id>\d+)/sub-events/(?P<guid>[\d\w-]+)/$' % urlpath, views.register_child_events, name='event.register_child_events'),
> ```
> 
> Then, in views.py,
> 
> ```
> def register_child_events(request, registration_id, guid=None,  template_name="events/reg8n/register_child_events.html"):
>     registration = get_object_or_404(Registration, pk=registration_id)
> 
>     perms = (
>         has_perm(request.user, 'events.change_registration', registration),  # has perm
>         request.user.is_authenticated and request.user == registration.registrant.user,  # main registrant
>         registration.guid == guid, 
>     )
> ```

updated!

PR #1215, opened 2023-09-08
Remove "coding: utf-8" file headers

BenSturmfels:
These headers are not required in Python 3 as the encoding is assumed to be UTF-8 by default.

Update was applied automatically with:

find . -type f -name '*.py' -exec perl -i -p -e 's/# -\*- coding: utf-8 -\*-\n//g' {} \;


jennyq:
Cool. Thanks @BenSturmfels !

PR #1217, opened 2023-09-09
Remove remaining Python 2 "u" string prefixes

BenSturmfels:
These changes were made automatically, with `pyupgrade`, but then trimmed down to include only the unicode literals-related changes for easier review:

  `git ls-files '*.py' | xargs pyupgrade --py3-only`

I've also included some changes that remove backwards-compatibility imports like:

  `from builtins import str`


BenSturmfels:
@jennyq these sorts of far reaching changes can be difficult to accept confidently without a lot of review, so alternatively, you could just run `git ls-files '*.py' | xargs pyupgrade --py3-only` on your own repository and commit all the changes found.

I've used pyupgrade extensively and its been very reliable - no issues. There's possibly less chance of issues from accepting its changes directly than from me sending PRs ;).

jennyq:
@BenSturmfels, 

Thank you! This is much needed update.  I think it's probably better to go with the alternative way you pointed out. Otherwise, I'm afraid it would cause a lot of merge conflicts with several other PRs we currently have.

BenSturmfels:
@jennyq that sounds good. I see that `setup.py` indicates you only support back to Python 3.7, so you may want to use:

`git ls-files '*.py' | xargs pyupgrade --py37-plus`

PR #1225, opened 2023-10-02
Print badges side by side instead of front and back

ssimmons42:


tendenci/apps/events/models.py
```diff
@@ -174,6 +174,9 @@ class Place(models.Model):
     country = models.CharField(max_length=150, blank=True)
     national = models.BooleanField(default=False, help_text=_('Is it a national event?'))
 
+    use_zoom_integration = models.BooleanField(default=False)
```
jennyq, line 177: These two fields, `use_zoom_integration` and `is_zoom_webinar`, are duplicates in the Place model.

jennyq:
Looks like badges are all packed together without spaces or separators . 

image

rschipul:
Jenny,

Can you send me the link to the sample event you are trying to print from?
I will test with the actual badge paper

PR #1230, opened 2023-11-02
Fix repeat of bug

ssimmons42:
Fix bug causing repeat_of to crash. repeat_uuid is needed to identify the original Event of the repeats. 
There will need to be a follow up to manually fix affected Events.

### DIFF
--- a/tendenci/apps/events/forms.py
+++ b/tendenci/apps/events/forms.py
@@ -965,15 +965,20 @@ class EventForm(TendenciBaseForm):
             #self.fields['groups'].initial = Group.objects.get_or_create_default()
             
         if 'repeat_of' in self.fields:
-            if self.instance.pk and self.instance.parent: # on event edit
-                if self.instance.repeat_of: # repeat_of already set
-                    self.fields['repeat_of'].queryset = Event.objects.filter(
-                        pk=self.instance.repeat_of.pk)
-                else:
-                    self.fields['repeat_of'].queryset = Event.objects.filter(parent_id=self.instance.parent.id).order_by('repeat_uuid').distinct('repeat_uuid')
-            elif 'parent' in self.fields and parent_event_id: # on sub-event add
-                parent_event = Event.objects.get(id=parent_event_id)
-                self.fields['repeat_of'].queryset = parent_event.child_events.order_by('repeat_uuid').distinct('repeat_uuid')
+            if self.instance.pk and self.instance.parent and self.instance.repeat_of: # on event edit, repeat_of already set
+                self.fields['repeat_of'].queryset = Event.objects.filter(pk=self.instance.repeat_of.pk)
+            elif (self.instance.pk and self.instance.parent) or ('parent' in self.fields and parent_event_id): # on editing sub-event with no repeat, or adding new sub-event
+                # Populate repeat_of options with all child events that are not already repeats.
+                parent_event = self.instance.parent if self.instance.pk else None
+                if not parent_event:
+                    parent_event = Event.objects.get(id=parent_event_id)
+                queryset = parent_event.child_events.filter(repeat_of__isnull=True)
+
+                # If there's an instance, make sure it's not in the repeat_of options (don't want an Event to repeat itself)
+                if self.instance.pk:
+                    queryset = queryset.exclude(pk=self.instance.pk)
+                self.fields['repeat_of'].queryset = queryset
+
 
         if self.instance.image:
             self.fields['photo_upload'].help_text = '<input name="remove_photo" id="id_remove_photo" type="checkbox"/> Remove current image: <a target="_blank" href="/files/%s/">%s</a>' % (self.instance.image.pk, basename(self.instance.image.file.name))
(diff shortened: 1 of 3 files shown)


tendenci/apps/events/models.py
```diff
@@ -2338,6 +2339,11 @@ def __init__(self, *args, **kwargs):
         super(Event, self).__init__(*args, **kwargs)
         self.private_slug = self.private_slug or Event.make_slug()
 
+    @property
+    def title_with_event_code(self):
+        """Title prefixed with event_code"""
+        return f'{self.event_code} - {self.title}' if self.event_code else self.title
+
```
ssimmons42, line 2346: this should have gone in previous commit....

tendenci/apps/events/forms.py
```diff
@@ -2127,13 +2132,13 @@ def __init__(self, registrant, is_admin, *args, **kwargs):
 
         for index, start_dt in enumerate(sub_event_datetimes.keys()):
             child_events = upcoming_child_events.filter(start_dt=start_dt)
-            choices = [(event.pk, f'{event.event_code} - {event.title}' if event.event_code else event.title) for event in child_events if not event.at_capacity]
+            choices = [(event.pk, event.title_with_event_code) for event in child_events if not event.at_capacity]
             # Check if registrant already has selection. If so, make sure it's in choices
```
ssimmons42, line 2136: Property added in next commit. This is so that we don't have duplicate options when an event is already selected

tendenci/apps/events/forms.py
```diff
@@ -965,15 +965,20 @@ def __init__(self, *args, **kwargs):
...
-                    self.fields['repeat_of'].queryset = Event.objects.filter(
-                        pk=self.instance.repeat_of.pk)
-                else:
-                    self.fields['repeat_of'].queryset = Event.objects.filter(parent_id=self.instance.parent.id).order_by('repeat_uuid').distinct('repeat_uuid')
-            elif 'parent' in self.fields and parent_event_id: # on sub-event add
-                parent_event = Event.objects.get(id=parent_event_id)
-                self.fields['repeat_of'].queryset = parent_event.child_events.order_by('repeat_uuid').distinct('repeat_uuid')
+            if self.instance.pk and self.instance.parent and self.instance.repeat_of: # on event edit, repeat_of already set
+                self.fields['repeat_of'].queryset = Event.objects.filter(pk=self.instance.repeat_of.pk)
+            elif (self.instance.pk and self.instance.parent) or ('parent' in self.fields and parent_event_id): # on editing sub-event with no repeat, or adding new sub-event
```
ssimmons42, line 970: If there's an instnace and it has a parent & repeat_of is set, then limit drop down to the repeat_of it's set to

tendenci/apps/events/forms.py
```diff
@@ -965,15 +965,20 @@ def __init__(self, *args, **kwargs):
...
+                # Populate repeat_of options with all child events that are not already repeats.
+                parent_event = self.instance.parent if self.instance.pk else None
+                if not parent_event:
+                    parent_event = Event.objects.get(id=parent_event_id)
+                queryset = parent_event.child_events.filter(repeat_of__isnull=True)
```
ssimmons42, line 975: If there's an instance, then try to use instance.parent
If parent_event is still null, then query for it with parent_event_id
Then get a queryset of the parent's child events that don't have repeat_of set (these will be originals only)

tendenci/apps/events/forms.py
```diff
@@ -965,15 +965,20 @@ def __init__(self, *args, **kwargs):
...
+
+                # If there's an instance, make sure it's not in the repeat_of options (don't want an Event to repeat itself)
+                if self.instance.pk:
+                    queryset = queryset.exclude(pk=self.instance.pk)
+                self.fields['repeat_of'].queryset = queryset
+
```
ssimmons42, line 981: filter out the event itself... don't want the event to copy itself

tendenci/apps/events/models.py
```diff
@@ -1607,11 +1607,12 @@ def register_child_events(self, child_event_pks, is_admin=False):
...
 
-                current_event = self.registrantchildevent_set.filter(
+                # This is the orignal event with a matching repeat_uuid that the user is registered for.
+                original_event = self.registrantchildevent_set.filter(
                     child_event__repeat_uuid=event.repeat_uuid).first().child_event
                 error = _(
                     f'{event.title} on {event.start_dt.date()} is a repeat of event on ' \
-                    f'{current_event.start_dt.date()}. Please select only one.')
+                    f'{original_event.start_dt.date()}. Please select only one.')
                 raise Exception(error)
```
ssimmons42, line 1616: original_event is closer in meaning to what we're doing here... it is the original event the user registered for that is a copy of an event the user is trying to register for

tendenci/apps/events/views.py
```diff
@@ -1608,7 +1608,7 @@ def get_child_events(request):
     parent_id = request.GET.get('parent')
     event = get_object_or_404(Event, id=parent_id)
 
-    child_events = event.child_events.order_by('repeat_uuid').distinct('repeat_uuid')
+    child_events = event.child_events.filter(repeat_of__isnull=True)
     options = [{'value': x.pk, 'text': x.title} for x in child_events]
```
ssimmons42, line 1612: Still filters out the repeats... but makes sure we have the originals

ndavis2 (approved):

ndavis2:
@eschipul introducing you to @ssimmons42 ! Thanks for helping review!

lashirley21:
Awesome and welcome @ssimmons42 ! 🩵

PR #1231, opened 2023-11-03
Fix Events missing repeat_uuid, and tie originals back to repeats

ssimmons42:
Script to fix existing Events that are missing `repeat_uuid` and to add the corresponding `repeat_uuid` back to originals that have repeats.


ssimmons42:
Tested manually on my dev environment with these steps:
1. Manually set all repeat_uuids on Events to None 
2.  Ran migration script
3.  Verified all Events had a repeat_uuid and that repeated Events were tied to originals by having the same repeat_uuid

ssimmons42:
Closing as all current Events are correct. We can use this script if needed (aka if we run into issues with older Events being used as repeats)

PR #1232, opened 2023-11-03
Add test for repeat UUID

ssimmons42:
Test for Events  original -> repeat relationship (repeat_of gets set on repeated Event, and repeat_uuid is the same on both the original and the repeat)


requirements.txt
```diff
@@ -82,3 +82,5 @@ pyzoom>=1.0.3
 PyJWT>=2.8.0
 # Django model utils (FieldTracker)
 django-model-utils>=4.3.1
+# Create test data
+model-bakery==1.17.0
```
ssimmons42, line 86: Adds a library that will easily generate fake data for testing.

tendenci/apps/events/tests.py
```diff
@@ -104,3 +106,19 @@
...
 #             Addon.objects.get(title='Test Addon')
 #         logger.info('Complete.')
+
+class EventTest(TestCase):
+    def test_repeat_of_relationship(self):
+        original_event = baker.make('events.Event')
+        user = baker.make('auth.User')
+        # Add new event by setting 'repeat_of'
+        repeat_event = copy_event(original_event, user, set_repeat_of=True)
+        self.assertEqual(original_event.repeat_uuid, repeat_event.repeat_uuid)
```
ssimmons42, line 116: The first part of this text, adds a new Event & uses 'repeat_of' We should end up with the same repeat_uuid on both the original & the new, repeated event

tendenci/apps/events/tests.py
```diff
@@ -104,3 +106,19 @@
...
+
+        # Edit existing event by setting 'repeat_of'
+        new_event = baker.make('events.Event')
+        self.assertNotEqual(original_event.repeat_uuid, new_event.repeat_uuid)
+        another_repeat_event = copy_event(original_event, user, set_repeat_of=True, copy_to=new_event)
+    
+        self.assertEqual(original_event, another_repeat_event.repeat_of)
+        self.assertEqual(original_event.repeat_uuid, another_repeat_event.repeat_uuid)
```
ssimmons42, line 124: This part of the test sets 'repeat_of' on an existing Event and makes sure both the original and the repeat have the same 'repeat_uuid' even though they had different repeat_uuids before repeat_of was set.

ndavis2 (approved):
Thank you for adding testing here @ssimmons42

PR #1233, opened 2023-11-06
Adjust certificates to work in Chrome, Edge and Firefox

ssimmons42:


ndavis2 (dismissed):

jennyq:
Hi @ssimmons42,

For the single certificate, the format looks fine in FireFox, but still a bit out in Chrome (see the image below).

image

ssimmons42:
Interesting it is still looking like this on your machine. Definitely had it working on Chrome before... I wonder if it has something to do with differences in the certificate images used. The CSS is set so that image height should be 1in... but it is possible if the test images you are using are larger than the ones I used maybe it takes up slightly more room and is causing this issue?  Could you check the exact height that the images are being rendered? I would like to compare it to what I have. Ultimately, it looks like we will probably just need to move that section up slightly to accommodate Chrome... but i was curious as to what is causing the difference.

jennyq:
@ssimmons42, It's **not** about the image, but the item underneath the image where it is rendered as 3 lines instead of 2 lines. 
image

Btw, that's on the live site not on my machine.

ssimmons42:
> @ssimmons42, It's **not** about the image, but the item underneath the image where it is rendered as 3 lines instead of 2 lines. image
> 
> Btw, that's on the live site not on my machine.

Oh ok, that makes sense. I just wanted to make sure nothing weird was happening with the CSS.

ssimmons42:
@jennyq  Made another adjustment to the margins:

image

jennyq:
> @jennyq Made another adjustment to the margins:
> 
> image

Thanks! I'll do some tests on the live site.

Issue #1200, opened 2023-08-14
It is possible to add regions to the membership reports?


jalovick:
Hi,

We would like to add charts for our various membership levels to Tendenci covering each region. Is it possible to add an option to show the totals as is available now, and a dropdown to also select per-region?

Thanks

Jamie

jennyq:
Hi Jamie, 

Yes, it's possible. Which membership reports were you referring to?

jalovick:
Hi,

Any of the report with existing charts, and any others that don't require going into explorer:
- /memberships/reports/overview/
- /memberships/reports/members_stats/
- /memberships/reports/member_quick_list/ (adding membership number, status, and Email would be useful on this report as well, similar to the active membership report)
- /memberships/reports/active_members/
- /memberships/reports/renewed_members/
- /memberships/reports/expired_members/
- /memberships/reports/members_in_renewal_period/
- /memberships/reports/members_in_grace_period/
- /memberships/reports/members_over_time/
- /memberships/reports/active_members_ytd/
- /memberships/reports/members_ytd_type/
- /memberships/reports/join_summary
- /invoices/report/overview/
- /memberships/reports/members_donated/ (adding the fact they donated may be a useful flag for the member list reports)

There are probably a few more reports that it's worth adding it, particularly the corporate ones (we don't currently use that feature), but basically anything that lists users or stats on users, it would be good to filter by region.

Thanks

Jamie

rob-hills:
@jennyq I'd be happy to have a crack at this if you could give me some pointers.  Maybe an example of something similar and which template and/or python file(s) contain the logic for the example?

TIA,

Rob

rob-hills:
> Hi @rob-hills, Thank you for pitching in. I'll create an example when I get the chance.

Thanks @jennyq , I don't necessarily think I need you to create an example.  What I believe I need is just an overview of which program and template files make up a report and how they interact.

If you were to just pick one of the reports that Jamie listed above and list the template(s) and python files that are required to generate the content, hopefully I can build from there.  I understand the MVC concepts behind Django, I'm just struggling to find which elements are behind a specific report.

jennyq:
@rob-hills, 

[...]

PR #1247, opened 2023-12-31
Membership Report filtering by region

rob-hills:
Hi @jennyq ,

This pull request includes the building blocks for applying a Region filter to membership reports (as requested in #1200).

For now it has been implemented only in the Member Quicklist report but I have tried to structure it so it can be quickly and easily added to other membership reports as listed [in a comment in #1200](https://github.com/tendenci/tendenci/issues/1200#issuecomment-1678579270).

I'm happy to implement this in the other reports, but thought it best if you reviewed my code before I make all the other changes.

As a side note, there is some javascript in [tendenci/themes/t7-base/templates/memberships/search-region-form.html](templates/memberships/search-region-form.html) which is supposed to remove empty URL parameters from the form before it is submitted (to clear the ugly "?region_id=" parameter from the URL).  However, it doesn't work and my javascript debugging skills have so far been too rusty to fix this.  If you happen to have a couple of minutes sometime to look at this and suggest a fix (or an alternative way of doing thiis) I'd be most grateful!

Cheers, and Happy New Year,

Rob


jennyq:
Thanks @rob-hills ! I'll review it this week when I get a moment. 

Happy New Year!

jennyq:
@rob-hills, 

Your code works functionally, but I refactor the code to make sure the user input is validated, and it is more Django friendly. 

I created a form `RegionForm` with a field `region` which uses `ModelChoiceField` with a list of regions as the `queryset`. That way, if you call `region_form.is_valid()`, it will automatically validate if the user input is within the range of the list of regions. 

Also the JS is fixed.

jennyq (dismissed):

jennyq (approved):

rob-hills:
Thank you for reviewing and refactoring.  The new code is certainly much cleaner!

I'll apply similar changes to some of the other reports in #1200

PR #1281, opened 2024-05-23
Update checkin

ssimmons42:
Update check-in so that the user scanning in registrants doesn't have to select a session and click "confirm". 
These changes will allow this workflow:

1) The user selects 'Start Check In' from the menu of a sub-event. The user is now set up to scan registrants into this session.
2) The user scans as usual (The initial check-in is still to the main event. Subsequent scans will be into sessions if there are sessions)
3) If the registrant can't be checked in, the appropriate error will be displayed and a modal to switch check-in sessions will pop up. 
4) If reminders are enabled, then a reminder will appear after the specified time (in Event settings) that will provide a link to switch sessions if they want to.


jennyq:
@ssimmons42, 

When checking in for a child event, if a registrant hasn't checked in to the parent event, you'd have to scan twice, one for parent event, one for child event. I think you can eliminate the second scan by moving down this line (Ln 4331 in events/views.py) `should_check_in_to_sub_event = registrant.should_check_in_to_sub_event`  until the parent event check-in has done (right above the if statement `if should_check_in_to_sub_event:`).

ssimmons42:
> @ssimmons42,
> 
> When checking in for a child event, if a registrant hasn't checked in to the parent event, you'd have to scan twice, one for parent event, one for child event. I think you can eliminate the second scan by moving down this line (Ln 4331 in events/views.py) `should_check_in_to_sub_event = registrant.should_check_in_to_sub_event` until the parent event check-in has done (right above the if statement `if should_check_in_to_sub_event:`).

It is currently requiring an initial scan for the parent before allowing them to scan into a child event because I believe that is the workflow of the client that requested it... there's an initial scan when they come in the door followed by subsequent scans for each child event. However, now that you mention it, I'm not 100% sure this is the case since we haven't talked about their workflow in months. I will double-check to see if this is the case or if they only start scanning with the first child event.

jennyq:
> It is currently requiring an initial scan for the parent before allowing them to scan into a child event because I believe that is the workflow of the client that requested it...

Ok, then leave it as it is.

ssimmons42:
> > It is currently requiring an initial scan for the parent before allowing them to scan into a child event because I believe that is the workflow of the client that requested it...
> 
> Ok, then leave it as it is.

I did double-check with Nicole to see if my memory on how they are using this is correct. If she comes back and says it's not, it is an easy adjustment to make.

jennyq:
The dropdown on the switch session doesn't look obvious and format is out.  Is it possible to make it look better? Maybe use radio buttons instead of a drop down? If the list is too long, make it scrollable.  

image

ssimmons42:
> The dropdown on the switch session doesn't look obvious and format is out. Is it possible to make it look better? Maybe use radio buttons instead of a drop down? If the list is too long, make it scrollable.
> 
> image

It definitely doesn't look like that on my machine... interesting... I will see if I can duplicate the issue then attempt to fix it. I'm wondering if i have some styling related files somewhere on my machine that it is using...

jennyq:
I was testing on iPhone.

ssimmons42:
> I was testing on iPhone.

ok thanks!  I can see how it might have some styling issues there... I will give that a try... especially since they will be using their phones.

ssimmons42:
@jennyq Made some mobile-friendly adjustments and here's the modal on my phone:
image


I also added `btn-default` to correct the color of the button since I noticed the color difference when viewing on my phone as well.
Going to push this change up right now.

jennyq:
That looks better! Thanks @ssimmons42 !

Can we make the date time in the drop friendly? Instead of "05/29/2024 - 05/29/2024", can it display "Wed, May 29 • 4:00 PM" or similar?

Also, there is a weird reflex when I click the link "Do you want to switch sessions?". It tries to check in **again** for what it previously checked in. So, it shows the message saying ".... is already checked into ..."
image

ssimmons42:
> That looks better! Thanks @ssimmons42 !
> 
> Can we make the date time in the drop friendly? Instead of "05/29/2024 - 05/29/2024", can it display "Wed, May 29 • 4:00 PM" or similar?
> 
> Also, there is a weird reflex when I click the link "Do you want to switch sessions?". It tries to check in **again** for what it previously checked in. So, it shows the message saying ".... is already checked into ..." image

Yeah, given that these are all child events happening today, the 5/29/2024 part isn't very useful at all... I can also play with the switching feature and see how it might be done a little more smoothly. I will take a look & let you know.

jennyq:
Thanks for the update, Shandy!

One more thing - When none of the sub-events has been stored in the django session yet, the message doesn't reflect that and can be confusing. 

So, in this case, can you make adjustment to change the title "Switch Check-in Session?" to "Start Check-in Session?", and message "You are not set to check in registrants to any Event. Do you need to switch sessions?" to "You are not set to check in registrants to any Event. Do you want to start the check-in with a session?"? Remove the button "Continue with Session" and change the label for the button "Switch Session". 

image

ssimmons42:
> Thanks for the update, Shandy!
> 
> One more thing - When none of the sub-events has been stored in the django session yet, the message doesn't reflect that and can be confusing.
> 
> So, in this case, can you make adjustment to change the title "Switch Check-in Session?" to "Start Check-in Session?", and message "You are not set to check in registrants to any Event. Do you need to switch sessions?" to "You are not set to check in registrants to any Event. Do you want to start the check-in with a session?"? Remove the button "Continue with Session" and change the label for the button "Switch Session".
> 
> image

Good catch! That wording is a little confusing for that case.

ssimmons42:
@jennyq 

Updated!
image

jennyq:
Hi Shandy, 

When there is only one event to check in, the reminder "Do you want to switch sessions?" still pops up. Most likely there would be multiple events in one day, so, maybe also check and dismiss the reminder if the current event is the last one in the day?

ssimmons42:
> Hi Shandy,
> 
> When there is only one event to check in, the reminder "Do you want to switch sessions?" still pops up. Most likely there would be multiple events in one day, so, maybe also check and dismiss the reminder if the current event is the last one in the day?

Updated it! It will only show if there is a session available to switch to now.

jennyq:
Regarding the "Start Check in" link, when I click it on **Mac**, it shows "Ready to check in to ...", but...

Ideally, it should only show on mobile devices (phones and tablets). And it should display to those users with the permission to do the digital check-in.

ssimmons42:
> Regarding the "Start Check in" link, when I click it on **Mac**, it shows "Ready to check in to ...", but...
> 
> Ideally, it should only show on mobile devices (phones and tablets). And it should display to those users with the permission to do the digital check-in.

Updated! It now only shows up on mobile... also added a check to make sure it's a super user.

jennyq (approved):

PR #1288, opened 2024-06-04
15.1 blank screen forums 403 HttpResponseForbidden should re-raise

CraigeHales:
when an authenticated user gets permission denied in tendenci/apps/forums/views.py, a blank page is returned because HttpResponseForbidden is called with no text. Replace return HttpResponseForbidden() in the RedirectToLoginMixin function with raise
to reraise the PermissionDenied exception works perfect.

It looks like this with the reraised exception. Fred (not a member) is trying to see a forum that needs a membership. Fred sees the correct message, not a blank screen.


jennyq:
Nice fix. Thanks @CraigeHales !

PR #1293, opened 2024-06-20
Prevent sending captcha for authorized user

CraigeHales:
send the current user to FormForForm so it can suppress the captcha for authorized users.
Many of the pages in the default Tendenci install have 2 or 3 forms that need a captcha for anonymous users. The captcha is supposed to be suppressed (I read somewhere...) when logged in. This change means there are 2 or 3 fewer captcha sent. The previous change was here: https://github.com/tendenci/tendenci/commit/60441a402c6aa32dab82fd429616b93f4ebbce46 about 13 years ago. My testing shows when not authenticated there is an AnonymousUser in context.request.user.  If there is another form of captcha, maybe from google, I have not tested that. 

---- the network tab on a browser shows the captcha ----
before
before

after
after

 @jennyq - I'm still stumbling around with github and github desktop; let me know if I am making unnecessary work, thanks!


jennyq:
Thanks @CraigeHales !

PR #1296, opened 2024-07-13
PYBB forum - use the Settings DateFormat in the topic list 

CraigeHales:
Make the date formats consistently use the value in global settings. 

**before** (the dates are ambiguous and not using the setting value)

dateBefore

**after** (the dates have the 3-letter month from capital M in the 3rd picture's date format)

dateAfter

**settings**

dateSettings


jennyq:
Thanks @CraigeHales !

PR #1302, opened 2024-08-04
Update 403 and 404 pages to include error codes

b-davies24:
We've had some very vague requests from members when there is a 403, so this is an update to 403.html and 404.html to include the error codes '403' and '404' to make our lives easier when diagnosing!


jennyq:
Thanks @b-davies24 !

Issue #1297, opened 2024-07-15
Remove pre-fill event locations from list


b-davies24:
When adding an event, there is the option to select a previously-entered location from a drop-down.
![image](https://github.com/user-attachments/assets/e8ca5673-f01b-4fe2-b8dc-41d022ad2745)
This is really useful, however, it will store ANY information that is entered in the location - we have multiple errors from when people have mistyped/changed spelling/etc.
![image](https://github.com/user-attachments/assets/0b24d9d6-3c65-459b-9b08-040efd7b3b80)
Upon investigating this, there seems to be no obvious way for an admin to modify or delete these.

How would you suggest we deal with this?

jennyq:
Currently there is no web interface to delete them. You can issue a sql query to delete them from db directly. For admin to easily manage places/locations, the Place model needs to be registered at admin backend. We'd accept a PR like that.

rob-hills:
@jennyq I think we'd need to merge duplicate entries and update the foreign keys in the events_event table or else we'll end up events that have no location!  Is the place table linked to any others?

jennyq:
> Is the place table linked to any others?

No, events_event is the only one it is linked to.

rob-hills:
Thanks for that @jennyq , I've been working on this and have something almost ready to submit in a PR.  I'm just waiting for some others to test it first.

PR #1306, opened 2024-08-16
Event place sorting and model admin

rob-hills:
This PR is a fix for #1297
It adds an Admin interface for the event.place model, incorporating a facility to merge event places to help clean up data in related Event records.
It also sorts the Events form "Place"  combobox list in "name" order first (then address etc as before).  I believe this is more intuitive to the user than sorting by Address as the address isn't visible in the combobox list.


rob-hills:
@jennyq it would be great if you could check this out and incorporate it into the next Tendenci point release if you're happy with the changes.
TIA

jennyq:
Nice! Thanks @rob-hills !

Issue #1311, opened 2024-09-22
Editing a Directory doesn't update its expiry date


rob-hills:
When you add a new Directory, its expiration_dt field value is set to equal the activation_dt plus the requested_duration.

However, when you edit a Directory entry, the expiration_dt value is not updated.  Hence, edits to either the activation_dt or the requested_duration values don't change the expiration_dt as expected.

PR #1313, opened 2024-09-22
Update directory expiration date after directory edit

rob-hills:
Update directory.expiration_dt after editing a directory to realign to any changes to start date or duration

This addresses issue #1311

### DIFF
--- a/tendenci/apps/directories/views.py
+++ b/tendenci/apps/directories/views.py
@@ -316,7 +316,11 @@ def edit(request, id, form_class=DirectoryForm, template_name="directories/edit.
                     #directory.logo = None
             
             # set the expiration date in case activation date and/or requested_duration have changed
-            directory.expiration_dt = directory.activation_dt + timedelta(days=directory.requested_duration)
+            if directory.activation_dt and directory.requested_duration:
+                directory.expiration_dt = directory.activation_dt + timedelta(days=directory.requested_duration)
+            else:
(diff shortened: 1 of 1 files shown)


tendenci/apps/directories/views.py
```diff
@@ -314,6 +314,10 @@ def edit(request, id, form_class=DirectoryForm, template_name="directories/edit.
                 except IOError:
                     pass
                     #directory.logo = None
+            
+            # set the expiration date in case activation date and/or requested_duration have changed
+            directory.expiration_dt = directory.activation_dt + timedelta(days=directory.requested_duration)
```
jennyq, line 319: Please add an if statement to make sure "directory.activation_dt" is not None. Also, this should only be available to superuser as regular users will not be able to edit the requested duration.

jennyq: Also, please check if "directory.pricing" exists or not.

rob-hills: @jennyq I've put in checks for directory.activation_dt AND for directory.requested_duration.

I had a look at the directory.add method and there are way more business rules in that around pricing, require_payment etc than there are in the edit method.  It occurs to me that maybe those rules should be pulled out into a single "gatekeeper" method that is called by both the add() and edit() methods?

jennyq:
Thanks @rob-hills !

PR #1322, opened 2024-10-22
Django5 upgrade

bernd-wechner:
Minimalist changes I needed to get runserver operational with Django 5.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 future
 six
 configparser
-Django>=4.2,<5.0
+Django>=5.0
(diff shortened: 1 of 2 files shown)


jennyq:
Hi @bernd-wechner, 

Thank you for the upgrade! However, as a best practice, we only upgrade Django when its latest LTS version comes out. In this case, we'll upgrade when Django 5.2 LTS is released.  

I can create a new branch django5, and merge this PR to it. Would you like me to do that?

bernd-wechner:
Fine by me. Thanks.

PR #1323, opened 2024-10-22
USER_IS_STAFF for reports seems more reasonable than USER_IS_SUPERUSER

bernd-wechner:
In the ideal, this would be configurable. But too many things in tendenci are only available to website admins (super users) and association reports clearly should be available to committee members and/or staff etc. At least, in any Association I've worked it. They don't all need or want to be website super users/admins.

### DIFF
--- a/tendenci/themes/t7-base/templates/top_menu/admin_top-new.html
+++ b/tendenci/themes/t7-base/templates/top_menu/admin_top-new.html
@@ -25,7 +25,7 @@
 
                 {% include 'top_menu/_apps_dropdown.html' %}
 
-                {% if USER_IS_SUPERUSER %}
+                {% if USER_IS_STAFF %}
                 {% include 'top_menu/_reports_dropdown.html' %}
                 {% endif %}
 


jennyq:
First of all, the`USER_IS_STAFF` variable is not included in the template context.

Secondly, the majority of the reports under the Reports tab are not accessible by non-superusers. 

I would suggest to just create a page to list all reports that accessible for your staffs.

bernd-wechner:
Personally, I think Tendenci relies far far too heavily on USER_IS_SUPERUSER. Also puzzles me that USER_IS_STAFF seemed to work for us, and that it has different template scope (you're suggesting) to USER_IS_SUPERUSER. That's a surprise I admit. 

There's a principle at play here that we have recently broken out of frustration: It is the precautionary principle, do not assign SUPERUSER to anyone who doesn't need it. The converse of that principle is to make sure things staff need to see like membership reports in our case, are available to them - I have to make the superusers to see these reports. A bad move in principle, but one I've been cornered into.

PR #1327, opened 2024-10-27
disable_template_cache was crashing.

bernd-wechner:
Turns out if settings are loaded more than once this one liner breaks down fast. Added a check that we're actually disabling the cached loader which is more robust and clearer to the reader as well.


jennyq:
Thanks @bernd-wechner !

Issue #1318, opened 2024-10-06
TinyMCE issues and top bar issues (maybe linked)


b-davies24:
We have discovered a few bugs, which seem to all be intertwined:

1. When exiting the fullscreen view of TinyMCE, particularly in Google Chrome, you will sometimes be directed to the homepage (or Dashboard, if signed in as Admin). This is reproducible in Firefox, but as mentioned in 2), the hyperlink is higher since the file/edit/insert/... bar is visible. We have tested this on multiple Chrome installations on different devices.

We have managed to reproduce this issue on the Tendenci Demo site - if you go to fullscreen on TinyMCE while editing a page/event, and click just at the top of the "File" menu, it will go to the /dashboard:
![image](https://github.com/user-attachments/assets/f7ab6c14-ea47-47e5-97e0-94b115fe7093)

It seems that the Tendenci logo in the grey bar is somehow coming through to TinyMCE, and the hyperlink is overlaying on some bits (more in 3)).

2. When in the fullscreen view of TinyMCE on Chrome, the grey menu of Tendenci blocks the file/edit/insert/... buttons at the top of TinyMCE. This seems to exacerbate the issue in 1)
![image](https://github.com/user-attachments/assets/ca74b4f9-c035-4b3a-9209-82d9a182d40e)

3. On deeper visual inspection, it seems the hyperlink for the Tendenci logo on the grey logged-in bar creeps down below the bar (both on our site and the Demo site) all the time - you think you're clicking the website logo, but you're actually clicking the Tendenci logo in the bar if (unless you click low enough).
![image](https://github.com/user-attachments/assets/f2cc8d43-aa7d-429d-a578-2b795e368d2e)
![image](https://github.com/user-attachments/assets/8581d5b0-8729-459e-87af-95dd232f9db3)

b-davies24:
After a very quick investigation, the height of the link for the image seems to extend below the bar: 
![image](https://github.com/user-attachments/assets/2354a227-5e74-4276-a640-303b166ddd2e)

This may be the location, where you could add `style="height: 27px;"` to override the CSS height of 65px:
https://github.com/tendenci/tendenci/blob/af0cc537a209384623b0cae2b6093e7867926c90/tendenci/themes/t7-base/templates/header.html#L9C13-L9C64

b-davies24:
See PR #1331 

Note: The issue with Chrome and TinyMCE is still there 
[...]

PR #1331, opened 2024-11-15
Update header.html; change height of t-main-navbar-brand

b-davies24:
The quick addition of setting height=27px as per part of issue #1318

This _may_ fix it, but I haven't tested this when in the template (I manually added it into the browser inspector). 

There may be a more elegant solution, but this should fix the link 'size' being too big for now.


jennyq:
Thanks @b-davies24!

This commit https://github.com/tendenci/tendenci/commit/0945f578ef8c4e6f6958cff0cb4e88e7a9155f80 should fix the issue.

PR #1338, opened 2025-02-16
Changed "Submit" to "Complete Registration" on register_child_events.html

b-davies24:
Also, the button was changed from default to success to be in line with other submit buttons.

This name/colour change makes it clearer to the user when going through the process of registration.


jennyq:
Thanks @b-davies24 !

Issue #1005, opened 2021-04-02
Phone Number Formatting


CWalkden:
In the Profiles module the phone number is displayed to a user in the format: (XXX) XXX-XXXX
In Australia the format should be either: XX XXXX XXXX (for landline) or: XXXX XXX XXX (for mobile).  There are other variations as well, but these would be my preference.
There are some localisation settings in Tendenci, is it possible to apply this to this field?

jennyq:
Yes, it's possible. I'm not sure which setting can be used. Maybe just use a new setting? Then the [phonenumber](https://github.com/tendenci/tendenci/blob/master/tendenci/apps/base/templatetags/base_filters.py#L458) filter can be updated based on the setting. You're welcome to submit a pull request.

rob-hills:
@jennyq , would you be open to modifying the [phonenumber](https://github.com/tendenci/tendenci/blob/master/tendenci/apps/base/templatetags/base_filters.py#L458) filter to use the [PyPi Phonenumbers](https://pypi.org/project/phonenumbers/) Library?  I would be happy to have a go at doing a pull request for this.

jennyq:
The python phonenumbers looks like a good library. My only concern is the memory overhead. Maybe just use [phonenumberslite](https://pypi.org/project/phonenumberslite/) version as they suggested. @eschipul, what do you think?

eschipul:
@jennyq I agree that as @CWalkden pointed out, given most of our client base is international, properly displaying the phone number is important. If the phonenumberslite module works to keep it fast, then that seems reasonable.

It might also be that if we are storing the phone number in a char field that we would only need to load either library on add/edit screens - and not displays, views or reporting.

jennyq:
Thanks @eschipul ! @rob-hills You're welcome to proceed with [phonenumberslite](https://pypi.org/project/phonenumberslite/).

rob-hills:
@jennyq I've created a phonenumberslite branch in our AEVA fork of the Tendenci code and done some basic testing.  I had originally put a couple of new configuration settings into the settings.py file but our other website admins wanted them in apps/site_settings/settings.json file instead so it could be modified from the Web Admin interface.
I've put my new settings into the AEVA fork apps/site_settings/settings.json file and installed onto our test server from my phonenumberslite branch the way I do normally when testing new code.  However the new settings aren't present on the settings/site/global webpage of our site.  Is there a special trick to deploy changes to the global site_settings settings.json file on a test website?

[...]

PR #1340, opened 2025-02-16
Use Phonenumberslite library for phone number formatting

rob-hills:
The changes in this pull request address issue #1005 and enable localisation of phone number parsing and display.

It uses the [phonenumberslite library](https://pypi.org/project/phonenumberslite/) for phone number input parsing.

Two new site settings are also added:
  phone_number_region
  phone_number_pattern

The phone_number_region setting allows the administrator to set their phone number region (ISO 3661-1) which modifies how the phonenumberslite interprets input phone numbers.

The phone_number_pattern setting gives admins control over how phone numbers are displayed.


jennyq:
Thank you @rob-hills !

rob-hills:
> Thank you @rob-hills !

You're welcome @jennyq !

> Should the +country_code part not to show if the original value of the phone number doesn't include the country code?

For sites with international members I think the +country_code should always display to minimise ambiguity.  Even primarily national sites like ours have some international members, or expats living overseas so again I think there's good reason to mandate the country code.
 
> Where to get a list of phone number regions for site admins to choose?

There's a hint in the description for the phone_number_region setting: "ISO 3116-1".    Conveniently there's a list [published on Wikipedia](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) which turns up pretty readily in a Google Search for ISO 3116-1, but perhaps I should put the link in that description as well? Is it possible to put a hyperlink in the description of a setting in the json file?  If so, do I put it in as a html `<a href... >` tag or use markdown or something else?

> Certainly, for the sites with international user base, parsing phone numbers based on the setting "Phone Number Region" probably would not work. Not sure the solution at the moment.

I've just tested this.  Our test system has AU set as the phone number region but if I enter the US IDD code (+1) it accepts that happily and stores and displays it instead of the Australian IDD code (+64).  So I think this is not a problem.

jennyq:
> > Should the +country_code part not to show if the original value of the phone number doesn't include the country code?
> 
> For sites with international members I think the +country_code should always display to minimise ambiguity. Even primarily national sites like ours have some international members, or expats living overseas so again I think there's good reason to mandate the country code.

Fair enough, we can keep it on.

> Is it possible to put a hyperlink in the description of a setting in the json file? If so, do I put it in as a html <a href... > tag or use markdown or something else?

Yes, it'd great if you can add a link in the description. You can take a look at how other settings include the links, for example, this one https://github.com/tendenci/tendenci/blob/main/tendenci/apps/site_settings/settings.json#L735. Basically, use single quotes or escape the double quotes. 

> I've just tested this. Our test system has AU set as the phone number region but if I enter the US IDD code (+1) it accepts that happily and stores and displays it instead of the Australian IDD code (+64). So I think this is not a problem.

Hmm, looks like the possible issue is **not** at "Phone Number Region", **but** "Phone Number Pattern". 
For example, with the pattern "(###) ###-####", it will not work for a France number like +33 6 12 34 56 78.

rob-hills:
@jennyq 
> > Is it possible to put a hyperlink in the description of a setting in the json file? If so, do I put it in as a html  tag or use markdown or something else?
> 
> Yes, it'd great if you can add a link in the description. You can take a look at how other settings include the links, for example, this one https://github.com/tendenci/tendenci/blob/main/tendenci/apps/site_settings/settings.json#L735. Basically, use single quotes or escape the double quotes.

OK, I've done this and committed it to the branch

jennyq:
@rob-hills It complains this branch is out-of-date with the base branch. Can you please do a merge from main into this branch?

rob-hills:
Oops, your _main_ moved again, I've just synchronised again!  Fast-moving target to try and keep up with :-D

jennyq:
Thanks @rob-hills !

PR #1342, opened 2025-02-28
Added method to include raw files within the static directory. 

rockinrobstar:
This is needed to add base64 images to pdf's.


jennyq:
Thanks @rockinrobstar !

PR #1343, opened 2025-03-04
Add check that payment mode is online

AnttiMJohansson:
Added check that payment is expected to be made online.
Currently if online payment is not active and member makes renewal that is in pending state and then click on the renewal link again, system launches to online payment that gives message of online payments not set up.
To address this, when expired membership has submitted renewal, we trigger online payment only if membership is paid online.


jennyq:
Thanks @AnttiMJohansson !

Issue #1345, opened 2025-03-11
Discount "End Date" column has incorrect label


rob-hills:
The Discount model's "End Date" column has the label "Start Date" instead of "End Date"

PR #1346, opened 2025-03-11
Correct Discount.end_dt column label

rob-hills:
Fix for #1345
Change "end_dt" column label from "Start Date" to "End Date"


jennyq:
Thanks @rob-hills !

Issue #1336, opened 2025-01-19
Events Log grows very large - does it need a maintenance interface?


rob-hills:
Today I discovered that our event_logs_eventlog has grown quite large: 17GB.

I ran the following Postgres query:

```
select
  table_name,
  pg_size_pretty(pg_total_relation_size(quote_ident(table_name))),
  pg_total_relation_size(quote_ident(table_name))
from information_schema.tables
where table_schema = 'public'
order by 3 desc;
```

Which produced the following results:
```
                      table_name                        | pg_size_pretty | pg_total_relation_size 
---------------------------------------------------------+----------------+------------------------
 event_logs_eventlog                                     | 17 GB          |            18614460416
 versions_version                                        | 239 MB         |              251027456
 notifications_noticeemail                               | 61 MB          |               64159744
 handler404_report404                                    | 35 MB          |               36544512
 django_session                                          | 15 MB          |               15589376
 emails_email                                            | 9232 kB        |                9453568
```

Clearly the event_logs_eventlog table footprint is vastly bigger than the next largest table.  Should this table have some sort of maintenance facility?  For example some admins might like to limit the amount of time they keep event_logs_eventlog data for.

Ideally though, there could be a regular process that summarises historical data and stores the summaries and deletes the individual records.

jennyq:
Nice idea! However, summarizing the historical data from event logs can be a daunting task.

rob-hills:
This issue took down our server again today so I need to sort it out.

@jennyq I am thinking of a fairly simple change:

- A setting in apps/site_settings/settings.json to define the number of months of EventLog history to retain, say `event_log_months_to_keep`;
- A new Event Log management command that runs nightly to delete all `event_logs_eventlog` records older than the `event_log_months_to_keep` number of months.

I would envisage that if `event_log_months_to_keep` is zero then no event_logs_eventlog records would be deleted. For backwards compatibility, zero would be the default setting.

If you're happy with this idea, I'll go ahead and code it up and then submit a Pull Request

jennyq:
@rob-hills Sounds good!

For the setting, it might be better to add a setting in conf/settings.py instead of a site setting in apps/site_settings/settings.json so that it won't be easily updated and accidentally delete all event logs. Thoughts?

Feel free to move forward. Thanks!

rob-hills:
[...]

PR #1350, opened 2025-03-22
Event log manager

rob-hills:
@jennyq this pull request is for a simple solution to #1336.

It adds a setting to global site settings ("keep_event_log_for_days") that enables site administrators to specify a number of days they wish to keep EventLog records for.

It adds a management command that deletes all records older than the specified number of days.

The default setting for keep_event_log_for_days (0) disables this functionality which maintains backwards compatibility.

I have also taken the liberty of adding this management command to the nightly_commands list

I have **not** at this time tried to implement any sort of statistics accumulation and I don't think there would be much call for this anyway.

### DIFF
--- a/tendenci/apps/base/management/commands/run_nightly_commands.py
+++ b/tendenci/apps/base/management/commands/run_nightly_commands.py
@@ -31,6 +31,7 @@ class Command(BaseCommand):
                     'cleanup_expired_dbdumps',
                     'clearsessions',
                     'make_recurring_payment_transactions',
+                    'prune_event_log_records',
                     )
 
         if get_setting('module', 'chapters', 'membershipsenabled'):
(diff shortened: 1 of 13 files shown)


jennyq:
Looks good! 

One small request - 
Can you please move the setting to settings.py? This is because the issue regarding "too much space taken by event logs" is not a real concern for site admins but DevOps. On the contrary, they might accidentally/unknowingly set some number that causes unintentional deletions.

rob-hills:
> Looks good!
> 
> One small request - Can you please move the setting to settings.py? This is because the issue regarding "too much space taken by event logs" is not a real concern for site admins but DevOps. On the contrary, they might accidentally/unknowingly set some number that causes unintentional deletions.

@jennyq OK, I've made these changes

jennyq:
Thanks @rob-hills !

PR #1363, opened 2025-05-30
 Convert span elements with datetime attributes to time elements

tristanjl:
datetime is an attribute of time - https://www.w3schools.com/Tags/tag_time.asp

This is resulting in "Date/time not in ISO 8601 format in field 'startDate'" from google's use of the event markup.


jennyq:
Thanks @tristanjl !

PR #1364, opened 2025-06-02
Add support for Python 3.13, dropping support for Python 3.8.

rockinrobstar:
Python 3.8 is EOL. https://devguide.python.org/versions/
Python 3..13 is declared in toml file however has errors when installing and running. These are fixed by this patch.

Update Pillow to 11.2.1 (latest), version 11.0 added support for Python 3.13 but removed 3.8 compatibility: https://pillow.readthedocs.io/en/stable/releasenotes/11.0.0.html

Update gevent to 25.5.1 (latest). Supports Python 3.8 and up. https://pypi.org/project/gevent/25.5.1/

Upgrade psycopg2 to 2.9.10. 

Fixed apps/registry/base.py to include new Metaclass types __firstlineno__ and __static_attributes__

Installs and runs clean on Linux (Fedora 41) aarch64.

TODO: Replace imghdr which is now depreciated. Temp fix by 'pip install standard-imghdr'
TODO: At some point we will need to consider move to psycopg 3 branch as psycopg2 is bugfix only.


jennyq:
Let's add this breaking change to the next major version T16. 
Thanks @rockinrobstar !

PR #1366, opened 2025-06-07
Add registration_open filter for event_list

tristanjl:
Add registration_open filter for event_list to allow list to select for only events that can currently be registered.


jennyq:
Thanks @tristanjl !

Issue #20, opened 2010-06-15
Articles-ShowH not implemented on several modules


aahmed60:
We need to implement showh on the following modules

Articles
News
Photos
Pages
Files
Settings (double check if this is required)
Entities (double check if this is required)

Please copy from T4

Issue #13, opened 2010-06-14
Photos-How do I navigate from the bulk edit page


aahmed60:
When I'm in the photos bulk edit page, I don't have an easy way back to the dashboard or the photos module.

PR #1379, opened 2025-08-14
Update base.html Juniper Theme welcome message

lashirley21:
I believe we should maintain our unique open source happy spirit. Over the years, we have received great feedback when developers read these lines, I know they make me happy.


tendenci/themes/juniper/templates/base.html
```diff
@@ -10,7 +10,14 @@
   <head>
     <meta charset="UTF-8">
     <!-- Juniper Theme Powered by Tendenci — The Open Source AMS
-    Fork us at github.com/tendenci  -->
```
lashirley21, line 13: I noticed I left line #13 - this repeats in line #20 where it should. Not sure how to edit but leaving comment :)

lashirley21 (commented):
I noticed  line #13 - this repeats in line #20 where it should. Not sure how to edit but leaving comment :) Jens

Issue #844, opened 2020-04-15
Hard Coded US Date format in a number of parts of the Code


rob-hills:
On our newly launched Tendenci-based site http://aeva.asn.au we have the Australian date format set in our Site settings (eg `'d-m-Y h:i A'` for the short date format).

However, our Membership manager noticed that his Memberships listing was showing US date format (https://www.example.com/admin/memberships/membershipdefault/).

We (@goetzk and I) looked for the source of this and discovered that the Memberships Views.py has   [two instances of a hard-coded US Date format](
https://github.com/tendenci/tendenci/blob/master/tendenci/apps/memberships/views.py#L1807).

We then did some grepping:
```
src/tendenci/tendenci/apps$ grep -rF "strftime('%m/%d/%Y')" --include=*.py .
```
which returned the following results:
```
./events/views.py:            process_options.append("--start_dt=%s" % start_dt.strftime('%m/%d/%Y'))
./events/views.py:            process_options.append("--end_dt=%s" % end_dt.strftime('%m/%d/%Y'))
./files/views.py:            'start_dt': start_dt.strftime('%m/%d/%Y'),
./files/views.py:            'end_dt': end_dt.strftime('%m/%d/%Y'),
./recurring_payments/models.py:                                           billing_cycle['start'].strftime('%m/%d/%Y'),
./recurring_payments/models.py:                                           billing_cycle['end'].strftime('%m/%d/%Y'))
./recurring_payments/models.py:                            self.billing_cycle_start_dt.strftime('%m/%d/%Y'),
./recurring_payments/models.py:                            self.billing_cycle_end_dt.strftime('%m/%d/%Y'))
./memberships/views.py:            'start_date': start_date.strftime('%m/%d/%Y'),
./memberships/views.py:            'end_date': end_date.strftime('%m/%d/%Y')})
./memberships/admin.py:            return dt.strftime('%m/%d/%Y')
./resumes/views.py:            start_dt = start_dt.strftime('%m/%d/%Y')
./resumes/views.py:            end_dt = end_dt.strftime('%m/%d/%Y')
```
So it appears that there are a number of locations that need to be inspected to see if the Site Settings date format should be applied instead.

rob-hills:
I should have mentioned, we upgraded to version 12.01 just prior to releasing our site!

jennyq:
Thank you Rob. Would you mind to submit a pull request?

rob-hills:
I'll see what I can do!

[...]

PR #1383, opened 2025-09-20
Hard coded dates - Core and Memberships app

rob-hills:
@jennyq this Pull Request is the first part of my work on hard-coded dates throughout Tendenci code and templates (Issue #844 ).  I've done it as two branches, "hard_coded_dates_core" and "hard_coded_dates_memberships" with the "memberships" branch being **branched off** the "core" branch.

The "Core" branch pulls the relevant Django settings into the Tendenci Settings file so they can be commented and made more visible to Tendenci web admins.  The date and time formatting settings have been given values that ensure they are as backwards-compatible as possible.  It also includes relevant changes to base code and templates.

The "Core" branch adds the Django "USE_TZ" setting and sets it to True (Django default is False).  Please note this will generate some warnings in the log files (`RuntimeWarning: DateTimeField received a naive datetime`) wherever "naive" DateTime objects are still in use.  These warnings will disappear as I work through the Tendenci code replacing DateTime objects with their Timezone equivalents.

Testing here (on https://test.aeva.asn.au ) has not thrown up any issues with these changes (apart from the log file warnings) so if you also don't find any major problems, I should be able to roll out similar fixes to the rest of the Tendenci code pretty quickly from now.

### DIFF
--- a/tendenci/apps/events/views.py
+++ b/tendenci/apps/events/views.py
@@ -3540,9 +3540,10 @@ def cancel_registration(request, event_id, registration_id, hash='', template_na
     if request.method == "POST":
         # If cancellation is no longer allowed, force a refresh so that
         # "cancellation not allowed message is displayed"
-        if not event.can_cancel:
-            return HttpResponseRedirect(reverse(
-                'event.cancel_registration', args=[event.pk, registration.pk]))
+        if not request.user.is_superuser:
+            if not event.can_cancel:
+                return HttpResponseRedirect(reverse(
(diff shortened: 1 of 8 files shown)


goetzk:
Great to see progress on #844!

I haven't looked at the code recently enough to know  off hand if things like Report() can take a differently formatted `start_date` and do the right thing so I can't comment on that sort of stuff.
I will note that there is a lot of (good) whitespace cleanup in the diff but they make it harder to pick out the actual changes. I'm not sure if your editor makes it possible to minimise those changes but it might be something which needs to be split out from the main code changes.

Issue #1136, opened 2022-11-13
Tax formatting for Australia


CWalkden:
Our organisation has grown to a size where we are required to charge tax on the things we sell.  Tax is to be included on memberships and other items that we sell, but not on donations.
In Australia, the tax is called GST (Goods & Services Tax), and is 10%.  Prices are typically quoted including GST, and then the GST is listed on the invoice as a component of the total.

Tendenci makes it a fiddle to get the GST showing correctly.  Here is a screenshot to demonstrate - this is from a form where we sell an item: [https://www.aeva.asn.au/forms/ev-licence-plate-tag/](url)

![image](https://user-images.githubusercontent.com/64681720/201544322-12b346f6-a881-460d-b780-7b60e35448ff.png)

We'd like to list the prices as $5.50 and $9.50, GST inclusive.  But the system only allows the prices to be listed excluding the tax.  This results in our odd-looking number $8.64.  We set this up in the form like so:
![image](https://user-images.githubusercontent.com/64681720/201544943-988509a0-35e8-4092-a3c9-3c9a8c541695.png)

Would it be possible to add an option to display prices including the tax component?

Similarly, in the memberships module:
![image](https://user-images.githubusercontent.com/64681720/201545300-6aed22f0-1ec3-46f5-9202-4d01c86c4465.png)

All prices exclude GST, but we'd prefer to have the membership prices displayed including GST.  The formatting would need to make this clear, here are some suggestions:  

- $55.00 (tax inclusive)
- $55.00 including GST
- $55.00 including $5.00 GST
- $55.00 optional text that we enter

goetzk:
Hi Chris,
Congrats on the growth :)

You'll probably also now encounter the issue in the invoices. We did some crazy stuff in the templates (which I won't share) but would also like the handling of tax to be improved.
IIRC I previously raised the 'Tax rate' dialogue but as your sceenshot shows, I didn't win that battle :)

jennyq:
Hi @CWalkden, 

For the custom forms, you can add the tax info to the "Pricing Description" field to inform user % tax will be applied.

For the memberships application forms, I've added the help text for the membership type field https://github.com/tendenci/tendenci/commit/68400a455d1f3d8af57c6d484e0d6857b17e2426.

I hope that addresses your issue. If not, a pull request is welcome.

[...]

Issue #1165, opened 2023-03-09
Another tax display problem


CWalkden:
I've created an Event with registrations and pricing.  On my pricing I selected the tax option.

For someone registering for the event, they first see the price on the event page:
![image](https://user-images.githubusercontent.com/64681720/223913765-b6ac13af-86ec-4e1d-ad98-56b71b3f5cb0.png)
No mention of tax here.
If they press the registration button, they get this:
![image](https://user-images.githubusercontent.com/64681720/223912374-4bc58529-42c6-496e-a06a-d8a1a083c990.png)
Normally in Australia, tax would be included in this amount.  So people are expecting to pay $10 at this point.

In the next section the tax is displayed like this:
![image](https://user-images.githubusercontent.com/64681720/223912183-2f963d84-7238-4517-a428-bc70b156be57.png)
Again, in Australia, this would mean that the $10 price includes $1 of tax.  But this is not the case.  When we progress to the invoice, we see the real situation:
![image](https://user-images.githubusercontent.com/64681720/223912959-7383ffae-f98b-40b8-90cf-465777759eca.png)

I think the first three views need to be changed so that they show the tax component.

jennyq:
Different regions may have their preferred ways to display tax. In US, tax is **not** included in the price of an item, people are expected to pay tax (if any) on checkout.  

Just added a "+" sign before "tax" on the confirmation page, so it would show "Price: $10.00 (+tax: $1.00)" https://github.com/tendenci/tendenci/commit/f919932fc12938b41eac0c138c2d9a2bcc01c39e. Does it help?

CWalkden:
> Just added a "+" sign before "tax" on the confirmation page, so it would show "Price: $10.00 (+tax: $1.00)" [f919932](https://github.com/tendenci/tendenci/commit/f919932fc12938b41eac0c138c2d9a2bcc01c39e). Does it help?

Yes, that definitely helps, thanks!

By the way, we also don't use the word 'tax'.  Here it is called 'GST'.  In the UK this is 'VAT', so this word might be something to consider as a variable for the website settings?

I did look at adding an explanation into the wording of the pricing, but it is used in too many places to work properly.  For instance "+ GST Light Meal" would work in the second screenshot, but not in the first one.


Ideally, the modules that use tax would give us the option to display tax included or excluded.  Then people could set it to suit their local preferences.  In some cases the way in which tax is displayed is law, so it is important to get right.

PR #1393, opened 2025-12-17
Add tax inclusive price option (EU/VAT/GST)

rockinrobstar:
Added option to have taxes included in prices (as is done in Europe and VAT and GST countries.

Fixed security flaw where donations set as negative will reduce invoice amount silently.

Fixed formatting on donations to show as a currency.


rockinrobstar:
Hi Jenny, please review my change. Hopefully I have caught all instances of where taxes are applied, but please let me know if I have missed any. 

Hopefully fixes: #1136 #1165

Issue #1396, opened 2026-01-29
Zoom participant names are Tendenci usernames


bje-:
When using the Zoom API integration for webinars, participant names are set to their Tendenci usernames rather than their full names. This means that it's hard for panelists to recognise "Joe User" if his participant name is `jxe655` (for example). It makes this quite hard to use.

PR #1397, opened 2026-01-29
Join Zoom meeting with the user's full name, not username

bje-:
This is a suggested (untested) fix for  #1396. Will this do the job?


jennyq:
That will not work if zoom credits need to be generated for the registrants.

This is because the username is a unique field, but first and last names are not. With username, you're guaranteed to find the exact same user on your site when generating zoom credits by polling users answers from zoom.

bje-:
I can investigate.

jennyq:
That sounds like a good workaround. You can implemented it with the second format, Ben Elliston | bje-. 

The code to generate zoom credits is here: https://github.com/tendenci/tendenci/blob/main/tendenci/apps/events/models.py#L3324.

PR #1413, opened 2026-05-10
Additional fixes for GST/VAT (tax included) countries

rockinrobstar:
Some additional fixes for GST/VAT tax treatment for tax inclusive countries that we have come across on our production system.


jennyq:
Thanks @rockinrobstar !

PR #1421, opened 2026-05-27
Improve message about the "Join" button appearing 10 minutes prior to a Zoom event.

### LINKED ISSUE #1419: Zoom 10 minute start notice is too small and not clear enough
Image

The text shown in this screenshot advising users that the meeting will become available 10 minutes prior to the start time is in too small of a font. The wording could also be a bit more explicit. I'd suggest:

```You will be able to join the Zoom meeting from 10 minutes before the start time of <START> until <END>.  Please come back then to find a "Join" button.```

bje-:
I left the `div` in place without any contents to avoid lots of reindenting. I think that's satisfactory.

Fixes #1419.


jennyq:
Thanks @bje- !

Issue #1408, opened 2026-04-22
Warnings: (templates.E003) 'static' is used for multiple template tag modules: 'django.templatetags.static', 'tendenci.apps.theme.templatetags.static'


rob-hills:
I've been getting alerts about this for a very long time.

```
System check identified some issues:

WARNINGS:
?: (templates.E003) 'static' is used for multiple template tag modules: 'django.templatetags.static', 'tendenci.apps.theme.templatetags.static'
```

I see this was previously raised in issue #1280 but the solution proposed was to suppress all template warnings and that issue was closed.  I'm not sure that's a good idea?

What's involved in fixing this?  I'm happy to have a crack at it if I can understand what it means.

[This answer](https://stackoverflow.com/a/71914347) to a similar question on Stack Overflow seems like it might be relevant?

jennyq:
Hi @rob-hills,

Tendenci has customized the static template tag and saved in "tendenci/apps/theme/templatetags/static.py". If you'd like to tackle this issue, here are the possible steps:

1) **Rename** the custom "static.py" to "tendenci_static.py".
2) **Update all of the references** `tendenci.apps.theme.templatetags.static` to `tendenci.apps.theme.templatetags.tendenci_static`.
3) If you look at the `TEMPLATES` setting in the settings.py, the custom static is added to the `builtins`, meaning there is no need to explicitly load the static tag with the code `{% load static %}` in any templates. After the static.py is renamed, there is no need to explicitly load the tendenci_static tag as well. However, some sites might have one or more templates pulled down with the tag "**{% load static %}**" in it. To make backwards compatible, those templates need to be updated to have "{% load static %}" removed. You can use **data migration** to handle that. 

Let me know if this proposed approach doesn't makes sense. Although the warning message is not urgent, it's kind annoying. Your help is much appreciated!

rob-hills:
Hi @jennyq , thanks for the details which I should be able to work with and I'm happy to have a go at fixing it.

I may need a little more guidance on using data migration for template updates.  Is there an example of this you can point me at?  Or documentation?

jennyq:
Regarding the data migration, first, you'd need to create an empty migration with the command:

    python manage.py makemigrations themes --empty

Then, you can add a function to handle the custom templates. Here is an example: https://github.com/tendenci/tendenci/blob/main/tendenci/apps/articles/migrations/0008_auto_20190816_1257.py.

rob-hills:
[...]

PR #1422, opened 2026-05-30
Rename templatetags static

rob-hills:
This PR fixes issue #1408 .

The code lists any template files that have been changed as part of the migration.


jennyq:
Nice job! Thanks @rob-hills !

PR #1426, opened 2026-06-22
Fix issues with event schema

tristanjl:
Currently, the event schema, when read by google, reports:

Invalid object type for field 'location'

This change addresses this error.


jennyq:
Thanks @tristanjl !

PR #1427, opened 2026-06-25
Only show pricing radio button if there is more than one price

tristanjl:
Currently, a radio button with no functionality is displayed if there is a single price for an event. This change makes it so radio buttons only appear if there is more than one option.


jennyq:
Thanks @tristanjl !

PR #1435, opened 2026-07-31
Cleanup whitespace

rob-hills:
@jennyq , as discussed, here is my pull request for purely whitespace clean up.

I have written a python routine that replaces all tab characters with four spaces and strips all whitespace from the end of each line.

I have done checks on a number of files to ensure that nothing else was changed except white space.  I've also run my local Tendenci app using the modified files and have not come across any issues.


jennyq:
Hi Rob,

The whitespace cleanup for the Python code looks good. 

However, it seems like too much for the HTMl templates. It's completely fine to have tabs in HTML, JS and CSS files. Can you please drop the changes for the HTMl templates, stylesheets and JavaScript files? Thanks!

rob-hills:
@jennyq I've reverted the commits that fixed the whitespace in Javascript, CSS and HTML files.

My code editor (Geany) doesn't discriminate at present, so I'll need to see if I can stop it automagically fixing whitespace in those files in future.

jennyq:
Thanks @rob-hills !

Issue #1439, opened 2026-08-20
Display Description field on user_group page


rob-hills:
Currently when a user/member views a User Group page, all they see is the name of the group.  This is not very helpful if we want them to subscribe to the group so it would be ideal if the Description field was also displayed.  This can then be used to provide a detailed description of the group and help interest more users in joining it/subscribing to it.

A simple PR will be submitted for this

PR #1440, opened 2026-08-20
Display user group description on Group View page

rob-hills:
Display the User Group Description so that, for example, users/members signing up for a mailing list group can see a description of what they are signing up for.

Submitted as a fix for issue #1439


jennyq:
Thanks @rob-hills !

